PR #28, opened 2019-07-02
Sprint-0/broadcast-on-turn

sibela46:


tillywoodfield (approved):

server/server.js
```diff
@@ -114,6 +114,18 @@ module.exports = port => {
         res.sendStatus(200);
     });
 
+    /* Check if it's my turn */
+    app.get('/api/myTurn', (req,res) => {
+        var user = currentUsers.find((user) => user.username === req.session.user);
```
Ro4052, line 119: One comment I would make is to try to avoid using `var` wherever possible. I would also favour `const` (or `let` if it needs to be mutable).

PR #40, opened 2019-07-03
Sprint 0/broadcast on turn

iatkins:


package.json
```diff
@@ -13,6 +13,7 @@
     "express-socket.io-session": "^1.3.5",
     "express-ws": "^4.0.0",
     "js-cookie": "^2.2.0",
+    "lint": "^0.7.0",
```
Ro4052, line 16: Was this package accidentally installed?

src/components/Dashboard/Dashboard.jsx
```diff
@@ -6,7 +6,7 @@ import Message from '../Message/Message'
 import PlayerCards from '../Cards/PlayerCards';
 import style from './Dashboard.module.css';
 import axios from 'axios';
-import { setGameState } from '../../store/actions';
+import { setGameState, setMessage } from '../../store/actions';
```
Ro4052, line 9: I know this isn't from this commit but you don't usally keep actions in the store folder. You end up with some nasty imports and it means the actions aren't close to the files they relate to.

src/components/Dashboard/Dashboard.jsx
```diff
@@ -17,11 +17,11 @@ class Dashboard extends React.Component {
...
     componentDidMount() {
         socket.on("gameState", msg => {
-            // this.setState({ gameState: msg });
-            console.log(msg);
-            
             this.props.setGameState(msg);
         });
+        socket.on("messages", msg => {
+            this.props.setMessage(msg);
+        });
```
Ro4052, line 24: I would consider moving the code in this method out of the component. It doesn't really belong here since it only interacts with redux and not the component itself.

src/components/Message/Message.jsx
```diff
@@ -50,7 +37,7 @@ class Message extends React.Component {
             <div>
                 {this.props.myTurn && box}
                 <div>
-                    <h1>{this.state.message}</h1>
+                    <h1 id="message">{this.props.message}</h1>
```
Ro4052, line 40: You don't actually need to give this an ID since you can search for an `h1` element in Enzyme. No harm in doing this, just pointing that out.

Ro4052 (commented):
Changes look good, most of the comments are more to do with existing code. Next steps I would say are looking at having separate reducer files for different parts of the app and moving socket code into a service file.

sibela46 (approved):

package.json
sibela46: Probably, I'll create an issue an remove it now

PR #47, opened 2019-07-04
Sprint 0/login tests

sibela46:


server/server.js
```diff
@@ -77,9 +77,9 @@ module.exports = port => {
             currentUsers.push(user);
             gameLogic.joinGame(user);
             emitGameState();
-            res.sendStatus(200);
+            res.status(200).json({message: "OK"});
```
Ro4052: I don't think you need to send the message OK since that's what the 200 code means.

server/server.js
```diff
@@ -77,9 +77,9 @@ module.exports = port => {
...
         } else {
-            res.status(404).json({message: "User already exists"});
+            res.status(200).json({message: "User already exists"});
```
Ro4052: Maybe user alread logged in?

src/components/Cards/PlayerCards.test.js
```diff
@@ -1,6 +1,17 @@
...
 import { PlayerCards } from './PlayerCards';
-import { shallow, mount } from 'enzyme';
+import { shallow } from 'enzyme';
+import moxios from 'moxios'
+import axios from 'axios';
+const express = require('express');
+const server = require('./mockserver');
+const request = require('supertest');
+
+const initServer = () => {
```
Ro4052: Is this used?

src/components/Cards/PlayerCards.test.js
```diff
@@ -41,3 +43,50 @@ describe('on given a set of cards', () => {
...
+
+describe('moxios', () => {
+    it('should install', () => {
+      let defaultAdapter = axios.defaults.adapter
+      moxios.install()
+      expect(axios.defaults.adapter === defaultAdapter).toEqual(false);
+      moxios.uninstall()
+    })
+
+    it('should uninstall', () => {
```
Ro4052, line 48: What are these testing?

src/components/Cards/mockserver.js
```diff
@@ -0,0 +1,17 @@
...
+    );
+    const {
+        cards
+    } = userData
+    return res.json({
+        cards
+    });
+  });
+  return router;
+};
```
Ro4052: The only function that uses this file looks like it's never being used?

src/components/Login/Login.jsx
```diff
@@ -16,21 +16,46 @@ export class Login extends React.Component {
...
+        var format = /[!@#$%^&*()_+\-=\[\]{};':"\\|,.<>\/?]+/;
+        if (format.test(string)){
+            return true;
+        } else {
+            return false;
+        }
+    }
+    checkUsername() {
+        if (this.state.value === '') {
+            window.alert('Please provide a username');
```
Ro4052: You might favour just not sending the response if there is an empty string. Could be a bit annoying for the user to have a window alert.

src/components/Login/Login.jsx
```diff
@@ -16,21 +16,46 @@ export class Login extends React.Component {
...
+            return false;
+        }
+        if (this.checkSpecialChar(this.state.value)) {
+            window.alert('Please provide a valid username')
```
Ro4052: Maybe look at erroring the input instead?

src/components/Login/Login.jsx
```diff
@@ -16,21 +16,46 @@ export class Login extends React.Component {
...
-            this.setState({
-                error: err.message
+        if (this.checkUsername()) {
+            axios.post('/auth/login', {
+                username: this.state.value
             })
-        })
+            .then(response => {                
+                if (response.data.message === "OK") {
+                    window.location = '/dashboard';
```
Ro4052, line 50: Take a look at `history.push`. You shouldn't be manually changing the window location like this.

server/server.js
Ro4052: Also, maybe it's best not to have a 200 response since technically shouldn't happen.

src/components/Login/Login.jsx
tillywoodfield: I've fixed this in my pr

src/components/Cards/PlayerCards.test.js
sibela46: If moxios set up correctly I think... I'm not really sure, I was just following a tutorial :D

tillywoodfield (approved):

src/components/Cards/PlayerCards.test.js
Ro4052: That's generally not the kind of thing you'd want to test, since it's not code that you've written.

PR #56, opened 2019-07-04
Redirect to main

sibela46:


src/components/Auth.jsx
```diff
@@ -0,0 +1,25 @@
...
+        }).catch((err) => {
+            console.log(err);
+            
+            if (window.location.pathname !== '/') {
+                window.location = '/';
+            }
+        })
+    }
+    render() {
+        return null;
```
Ro4052: I would move this into an action creator rather than a renderless component. Doesn't really make sense to have this as a component. You can have actions that are called as soon as the app starts. They don't have to be called from a component.

src/components/Login/Login.jsx
```diff
@@ -1,15 +1,36 @@
...
     }
+    componentDidMount() {
+        axios.get('/auth', {
+            validateStatus: (status) => {
+              return status < 500;
+            }})
+            .then(res => {            
+                if (res.status === 200) {
+                    this.setState({ loggedIn: true })
+                } else if (res.status === 400) {
```
tillywoodfield: Can we use 401 UNAUTHORISED for failed log in requests?

src/components/Login/Login.jsx
```diff
@@ -47,17 +68,23 @@ export class Login extends React.Component {
...
-                    window.location = '/dashboard';
+                    console.log(response);
+                    
+                    this.setState({
+                        loggedIn: true
+                    })
                 }
             })
             .catch(err => {                
                 if (err.message.includes(400)) {
```
tillywoodfield: 409 CONFLICT?

src/components/Login/LogoutButton.jsx
```diff
@@ -31,7 +37,9 @@ export class LogoutButton extends React.Component {
     }
     render() {        
         return (
+            (this.state.loggedIn) ?
```
tillywoodfield, line 40: I think this stuff can be removed if we're wrapping up the whole Dashboard component now?

sibela46: That's when a person is already logged in and needs to get redirected to dashboard

tillywoodfield (approved):

PR #62, opened 2019-07-05
Added some tests, to be improved

sibela46:


sibela46:
Ideally, all components' tests need to cover at least 80% of the code but I don't really know what other tests to write, so I'm creating an issue. In the future, we can make them all above 80...

EllieHield (dismissed):

src/components/Dashboard/Dashboard.test.js
```diff
@@ -31,26 +32,40 @@ describe('on initial render', () => {
...
+        const wrapper = shallow(<Dashboard currentWord={"Freedom"}/>);
+        expect(wrapper.find('#message').text()).toEqual("Word: Freedom");
+    });
     it("doesn't display the start game button", () => {
         const wrapper = shallow(<Dashboard status={firstRoundGameState.status} />);
         expect(wrapper.exists('#start-game')).toEqual(false);
     });
-});
-
-describe('after you have played your turn', () => {
```
tillywoodfield, line 51: Why was this test removed?

src/components/PlayerInfo/PlayerComponents.test.js
```diff
@@ -0,0 +1,92 @@
+import React from 'react';
```
tillywoodfield: Can this be split up so that each component has it's own test file?

src/components/Dashboard/Dashboard.test.js
sibela46: Because the end turn button sits in the PlayWordAndCard component, I thought I should move it in those tests

src/components/PlayerInfo/PlayerComponents.test.js
sibela46: Was thinking that, but thought the components are kinda related, so decided to have the tests in the same file. But I'll split it now.

src/components/Cards/AllCards.jsx
```diff
@@ -6,8 +6,7 @@ export class AllCards extends React.Component {
     getCardItems() {
         if (this.props.cards) {
             return this.props.cards.map(card => (
-                <img className={styles.allCards} key={card.id} src={require(`../Dashboard/dixit_logo.png`)}></img>
-                
+                <img className={styles.allCards} key={card.id} src={require(`../Dashboard/dixit_logo.png`)}></img>
```
tillywoodfield, line 9: Please can you add an alt here of the card id like in the other places just so React doesn't complain

tillywoodfield (approved):

PR #64, opened 2019-07-08
Sprint 0/others play card

tillywoodfield:


src/components/Cards/AllCards.jsx
```diff
@@ -3,29 +3,29 @@ import { connect } from 'react-redux';
 import styles from './Cards.module.css';
 
 export class AllCards extends React.Component {
+
     getCardItems() {
         if (this.props.cards) {
             return this.props.cards.map(card => (
-                <img className={styles.allCards} key={card.id} src={require(`../Dashboard/dixit_logo.png`)}></img>
+                <img className={styles.allCards} alt='' key={card.id} src={require(`../Dashboard/dixit_logo.png`)}></img>
```
Ro4052: Out of curiosity, does react complain if you don't have an alt on the image?

sibela46: Can the alt be the id of the card?

src/components/Cards/PlayedCards/PlayedCards.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import styles from '../Cards.module.css';
+
+export class PlayedCards extends React.Component {
+    render() {
+        return (
+            <ul id="played-cards">
+                {this.props.cards.map(card => {
+                    return <img id={"card-" + card.id} className={styles.allCards}  alt={"card-" + card.id} key={card.id} src={require(`../cards/card (${card.id}).jpg`)}/>
```
Ro4052: You could have this as a "one-liner", you don't need the return here.

src/components/Dashboard/Dashboard.jsx
```diff
@@ -33,14 +34,16 @@ export class Dashboard extends React.Component {
...
+        console.log("status ", this.props.status);
         return (
             <div className={style.roundInfo}>
                 {this.props.status === "NOT_STARTED" && <button id="start-game" onClick={this.startGame}>Start game</button>}
                 {this.props.status !== "NOT_STARTED" && <h2>Round: <span id="round-number">{this.props.roundNum}</span></h2>}
                 {this.props.currentPlayer && <h2>Current player: <span id="current-player">{this.props.currentPlayer.username}</span></h2>}
                 {this.props.currentWord !== '' && <h1 id="message">Word: {this.props.currentWord}</h1>}
                 <Players />
-                <AllCards />
+                {/* <AllCards /> */}
```
sibela46: If we're not using AllCards anymore, we can get rid of it

Ro4052 (commented):
Looking good.

src/components/Cards/PlayedCards/PlayedCards.jsx
sibela46: A Pure Component or? @Ro4052

src/components/Cards/AllCards.jsx
Ro4052: Usually the `alt` property describes the image for accessibility reasons. Although, with this game, I can't image it being very usable for people with sight problems... Think the `alt` prop could kind of break the game too.

sibela46 (commented):
Also, do you want to change all the action types to be constants and export them from the actions.js file (the way it's done for playerActions)

src/components/Cards/AllCards.jsx
tillywoodfield: Yeah I only added it in because React gives a warning if it doesn't have it

sibela46: @tillywoodfield

tillywoodfield: @Ro4052 what do you mean by it could break the game?

src/components/Cards/PlayedCards/PlayedCards.jsx
Ro4052: I meant just the map. I wouldn't bother making this a Pure component, if you want to optimise, you'd just make this a function component.

src/components/Cards/AllCards.jsx
Ro4052: So the game is based on coming up with a word that represents the image. The `alt` prop also represents the image. We don't want to be giving players hints like that.

sibela46 (dismissed):

sibela46 (approved):

PR #65, opened 2019-07-08
Sprint 0/disable after play card

tillywoodfield:


src/components/Cards/PlayerCards.jsx
```diff
@@ -51,9 +51,10 @@ export class PlayerCards extends React.Component {
         return ((this.props.playWordAndCard || this.props.playCard) && this.props.playedCard === 0) ? styles.singleCard : styles.disabledCard
     }
     render() {
+        console.log(this.props.playedCard);
         const cardsImages = this.getPlayerCards().map((card) => (
-            <img id={"card-" + card.id} alt='' className={this.disableOnEndTurn()} key={card.id} src={card.url} onClick={this.playCard.bind(this)}/>
-        ))
+            <img id={"card-" + card.id} alt='' className={this.disableOnEndTurn()} key={card.id} src={card.url} onClick={this.props.playedCard === 0 ? this.playCard.bind(this) : () => {}}/>
```
Ro4052: I'm not sure I like this line. I think if it were me, I'd bind the `playCard` function in the constructor. I'd also not pass the empty function in here but handle the case where `this.props.playedCard !== 0` in the `playCard` function.

tillywoodfield: We used to have it handled in the playCard function but ended up moving it so that in the tests we could see if playCard has been called, is there another way around this? Should we have one function for handling the click and then on the condition call the playCard function?

sibela46 (approved):

PR #72, opened 2019-07-08
Alsatian/server refactor

tillywoodfield:


server/services/gameLogic.js
```diff
@@ -1,34 +1,66 @@
...
+    cardId
+} */
 let votes = [];
 
-/* Return the list of players */
-exports.getAllPlayers = () => {
-    return allPlayers;
+/* Return the list of players, without their assigned cards */
+exports.getPlayers = () => {
+    return players.map(player => ({ username: player.username, score: player.score }));
```
Ro4052: Another one that can be a one-liner.

Ro4052: Same with other functions in this file.

sibela46 (dismissed):

server/services/gameLogic.js
```diff
@@ -51,91 +83,101 @@ exports.quitGame = player => {
...
+const playerHasVoted = (username) => {
+    return votes.filter(vote => vote.username === username).length > 0;
 }
 
 /* Returns true if all players have played a card this round */
 const allPlayersPlayedCard = () => {
-    for (let player of allPlayers.filter(player => currentPlayer !== player)) {
-        if (!player.playedCard) return false;
+    for (let player of players) {
+        if (!playerHasPlayedCard(player.username)) return false;
```
Ro4052: These two `for` loops in `allPlayersPlayedCard` and `allPlayersVoted` could be implemented using `every`.
`const allPlayersPlayedCard = () => players.every(player => playerHasPlayedCard(player))`
Something like that.

src/components/Cards/PlayedCards/PlayedCards.jsx
```diff
@@ -4,9 +4,10 @@ import styles from '../Cards.module.css';
 
 export class PlayedCards extends React.Component {
     render() {
+        console.log(this.props.cards);
         return (
             <ul id="played-cards">
-                {this.props.cards.map(card => <img id={"card-" + card.id} className={styles.allCards} alt={"card-" + card.id} key={card.id} src={require(`../cards/card (${card.id}).jpg`)}/>)}
+                {this.props.cards.map(card => <img id={"card-" + card.cardId} key={card.cardId} className={styles.allCards} alt={"card-" + card.cardId} key={card.id} src={require(`../cards/card (${card.cardId}).jpg`)}/>)}
```
Ro4052: I could be wrong but I don't think you actually need to `require` the image file here. You might be able to just use relative paths.

src/components/PlayerInfo/Players.jsx
```diff
@@ -8,7 +8,7 @@ export class Players extends React.PureComponent {
             <div className={style.currentPlayersBox}>
                 <h3>Players:</h3>
                 <ul id="players">
-                    {this.props.allPlayers.map((player, key) => {
+                    {this.props.players.map((player, key) => {
                         return <li key={key}>{player.username}</li>
```
Ro4052: OK, I'll stop pointing out unecessary returns now, you get the point.

sibela46 (approved):

PR #80, opened 2019-07-08
auth end to end test

iatkins:


sibela46 (commented):
What happens if an user is unauthorised? You can look into faking errors using cypress and maybe check that the Auth components deals correctly with the it (i.e. doesn't redirect to '/dashboard' and stays at '/'

sibela46 (approved):

PR #83, opened 2019-07-09
Alsatian/e2e dashboard

sibela46:


cypress/integration/dashboard_spec.js
```diff
@@ -1,3 +1,79 @@
...
+            });
+        })
+    
+        after(() => {
+            cy.request({
+                url: '/auth/logout',
+                method: 'POST'
+            });
+        })
+        it('successfully loads', function() {
```
Ro4052: I don't think we should be testing this. If the page doesn't load, we have a bigger problem.

cypress/integration/dashboard_spec.js
```diff
@@ -1,3 +1,79 @@
...
+        })
+
+        after(() => {
+            cy.request({
+                url: '/api/end',
+                method: 'GET'
+            });
+        })
+
+        it("after start has been pressed", () => {
```
Ro4052: May want to consider what we're using as test titles. This might be fine but it's not very descriptive.

server/services/gameLogic.js
```diff
@@ -38,7 +38,7 @@ let cards = [];
 let votes = [];
 
 /* Return the list of players, without their assigned cards */
-exports.getPlayers = () => players.map(player => ({ username: player.username, score: player.score }));
+const getPlayers = exports.getPlayers = () => players.map(player => ({ username: player.username, score: player.score }));
```
Ro4052: This is just a stylistic suggestion and you may well disagree.
```suggestion
const getPlayers = () => players.map(player => ({ username: player.username, score: player.score }));
exports.getPlayers = getPlayers;
```

src/components/Auth/Auth.jsx
```diff
@@ -10,20 +10,20 @@ export default class Auth extends React.Component {
       authenticated: false
     };
   }
-  componentDidMount() {
+  componentWillMount() {
```
Ro4052, line 13: `componentWillMount` is equivalent to the constructor. I think the general consensus is that this kind of thing would be called in `componentDidMount` but it doesn't really matter since API calls are asynchronus. However, will the API calls be moving into actions at some point?

src/components/Cards/PlayedCards/PlayedCards.jsx
```diff
@@ -7,7 +7,7 @@ export class PlayedCards extends React.Component {
         console.log(this.props.cards);
         return (
             <ul id="played-cards">
-                {this.props.cards.map(card => <img id={"card-" + card.cardId} key={card.cardId} className={styles.allCards} alt={"card-" + card.cardId} src={`../cards/card (${card.cardId}).jpg`}/>)}
+                {this.props.cards.map(card => <img id={"card-" + card.cardId} key={card.cardId} className={styles.allCards} alt={"card-" + card.cardId} src={require(`../cards/card (${card.cardId}).jpg`)}/>)}
```
Ro4052, line 10: Did it not work without the `require`?

src/components/Cards/PlayerCards.jsx
```diff
@@ -53,7 +53,7 @@ export class PlayerCards extends React.Component {
         return (
             <div className={styles.cardsContainer} id="my-cards">
                 <ul>
-                    {this.props.myCards.map((cardId) => <img id={"card-" + cardId} alt={"card-" + cardId} className={this.getCardClass()} key={cardId} src={`./cards/card (${cardId}).jpg`} onClick={this.handleClickCard} />)}
+                    {this.props.myCards.map((cardId) => <img id={"card-" + cardId} alt={"card-" + cardId} className={this.getCardClass()} key={cardId} src={require(`./cards/card (${cardId}).jpg`)} onClick={this.handleClickCard} />)}
```
Ro4052, line 56: Same as above.

src/components/Cards/PlayedCards/PlayedCards.jsx
sibela46: Nope

src/components/Auth/Auth.jsx
sibela46: I changed it to componentDidMount in the other PR... I think since we're not storing any state with the logging right now (the cookie is just saved in the browser), we don't need to move the authorisation requests to the actions for now (I spoke to Alex about this and that's what he told me)

Ro4052: OK, fair enough. I'm not sure I agree though, it would make sense that `loading` is a global app state and `authenticated` definitely seems like something that should be available outside of this component.

src/components/Cards/PlayedCards/PlayedCards.jsx
Ro4052: Ah well. It's not something I've done before but I thought I'd seen someone do it without `require` before.

src/components/Auth/Auth.jsx
sibela46: That's true, I could create an issue for it and we could fix it if we have time in this sprint?

src/components/Cards/PlayedCards/PlayedCards.jsx
tillywoodfield: It works without require only when you run the client and server separately, but I think when you build the front end if it doesn't have require then the string stays the same and it breaks cause the references are wrong

server/services/gameLogic.js
```diff
@@ -38,7 +38,8 @@ let cards = [];
 let votes = [];
 
 /* Return the list of players, without their assigned cards */
-exports.getPlayers = () => players.map(player => ({ username: player.username, score: player.score }));
+const getPlayers = () => players.map(player => ({ username: player.username, score: player.score }));
+exports.getPlayers = getPlayers;
```
tillywoodfield, line 42: This can be one line `const getPlayers = exports.getPlayers = () => players.map(player => ({ username: player.username, score: player.score }));`

server/services/gameLogic.js
```diff
@@ -87,6 +88,21 @@ exports.startGame = () => {
     }
 }
 
+/* Start the game with the players that have joined */
+exports.endGame = () => {
+    if (status !== statusTypes.NOT_STARTED) {
+        status = statusTypes.NOT_STARTED;
+        players = []
+        votes = []
+        card = []
```
tillywoodfield, line 97: Said this in the other pr too, but cards?

src/components/Dashboard/Dashboard.jsx
```diff
@@ -14,24 +14,20 @@ import PlayedCards from '../Cards/PlayedCards/PlayedCards';
 import axios from 'axios';
 
 export class Dashboard extends React.Component {
-
     constructor(props) {
         super(props);
         this.state = {}
     }
-
```
tillywoodfield, line 22: Something small but can you put this whitespace back in? I think it's pretty standard to have a line between each function in a React component

src/components/Auth/Auth.jsx
Ro4052: It's something that maybe warrants a discussion but yeah, don't worry about it for now.

src/components/Cards/PlayedCards/PlayedCards.jsx
Ro4052: Oh OK, that makes sense. The file structure would be different for the build. In that case, `require` will work best, since it can be used in both dev and prod.

server/services/gameLogic.js
Ro4052: Ironically, it was my comment that switched it from a one liner. I think this way looks a little cleaner but that's a personal preference.

tillywoodfield: Oh haha fair enough, I just saw someone do it online as a one-liner but I'm happy to do two if you think it's nicer

server/services/gameLogic.js
tillywoodfield: This still hasn't been fixed

src/services/socket.js
```diff
@@ -8,7 +8,7 @@ const connectSocket = () => {
     if (process.env.NODE_ENV === "development") {
         connectionString = "ws://localhost:8080";
     } else {
-        const wsProtocol = window.location.protocol === 'http' ? 'ws' : 'wss';
+        const wsProtocol = window.location.protocol === 'http' ? 'wss' : 'ws';
```
tillywoodfield, line 11: Wrong way round still?

Ro4052: Good catch.

PR #85, opened 2019-07-09
Alsatian/bugs/login

sibela46:


server/controllers/index.js
```diff
@@ -22,9 +22,13 @@ router.post('/auth/login', (req, res) => {
...
         };
-        currentUsers.push(user);
-        gameLogic.joinGame(user);
-        res.sendStatus(200);
+        if (gameLogic.canJoinGame()) {
+            currentUsers.push(user);
+            gameLogic.joinGame(user);
+            res.sendStatus(200);
+        } else {
+            res.sendStatus(403);
```
tillywoodfield: I think 403 is more for 'you don't have the permission level for this request'?

server/services/gameLogic.js
```diff
@@ -37,38 +37,48 @@ let cards = [];
 } */
 let votes = [];
 
-/* Return the list of players, without their assigned cards */
-exports.getPlayers = () => players.map(player => ({ username: player.username, score: player.score }));
```
tillywoodfield, line 41: Why have you removed this?

server/services/gameLogic.js
```diff
@@ -37,38 +37,48 @@ let cards = [];
...
+/* Get the current word of the game */
+exports.getWord = () => currentWord;
 
-/* Return the list of cards played this round, without who played them */
-const getCards = exports.getCards = () => cards.map(card => ({ cardId: card.cardId }));
```
tillywoodfield, line 44: Why have you removed this?

server/services/gameLogic.js
```diff
@@ -87,6 +97,21 @@ exports.startGame = () => {
     }
 }
 
+/* Start the game with the players that have joined */
+exports.endGame = () => {
+    if (status !== statusTypes.NOT_STARTED) {
+        status = statusTypes.NOT_STARTED;
+        players = []
+        votes = []
+        card = []
```
tillywoodfield: typo here, should be cards?

src/components/Auth/Auth.jsx
```diff
@@ -16,24 +16,23 @@ export default class Auth extends React.Component {
         return status < 500;
       }})
       .then(res => {            
-          if (res.status === 200) {
```
tillywoodfield, line 19: I think we're using 4 spaces tabs everywhere else, how come you've changed this file to 2?

src/services/socket.js
```diff
@@ -8,7 +8,7 @@ const connectSocket = () => {
     if (process.env.NODE_ENV === "development") {
         connectionString = "ws://localhost:8080";
     } else {
-        const wsProtocol = window.location.protocol === 'http' ? 'ws' : 'wss';
+        const wsProtocol = window.location.protocol === 'http' ? 'wss' : 'ws';
```
tillywoodfield: Isn't this the wrong way round?

server/services/gameLogic.js
sibela46: I haven't, it's just moved down

server/controllers/index.js
sibela46: https://httpstatusdogs.com/403-forbidden

server/services/socket.js
```diff
@@ -8,12 +8,30 @@ let sockets = [];
 exports.setupSocket = (server, session) => {
     io = socketio(server);
     io.use(sharedsession(session));
-    io.on('connection', function (socket) {
+    io.on('connection', function (socket) {
```
Ro4052: Well, I don't know what to say about this line : )

src/components/Auth/Auth.jsx
sibela46: Oh damn, it was when I copied it over from slack ahhah, sorry

src/services/socket.js
sibela46: it works for me

server/controllers/index.js
tillywoodfield: I dunno, I've seen different definitions in different places, on the Mozilla website it says:
403 Forbidden
The client does not have access rights to the content, i.e. they are unauthorized, so server is rejecting to give proper response. Unlike 401, the client's identity is known to the server.

tillywoodfield: For custom errors maybe we should just use a 400 bad request with a body containing the reason for the error? e.g. 400 "Error: the game has already started" @Ro4052 thoughts on status code use?

server/services/gameLogic.js
```diff
@@ -37,38 +37,52 @@ let cards = [];
...
+
+/* Get the current status of the game */
+exports.getCurrentPlayer = () => currentPlayer;
 
 /* Get the list of cards for a specific player */
-exports.getCardsByUsername = (username) => players.find(player => player.username === username).cards;
+exports.getCardsByUsername = (username) => {
+    let player = players.find(player => player.username === username)
+    if (player) return player.cards.filter(card => card.played === false)
+    else return []
```
tillywoodfield: We might run into problems later if we're hiding this error with an empty list if that makes sense?

server/controllers/index.js
```diff
@@ -21,10 +21,14 @@ router.post('/auth/login', (req, res) => {
         req.session.user = req.body.username;            
         const user = {
             username: req.body.username
-        };
-        currentUsers.push(user);
-        gameLogic.joinGame(user);
-        res.sendStatus(200);
+        };        
+        if (gameLogic.canJoinGame()) {
```
tillywoodfield: canJoinGame takes an argument

server/services/gameLogic.js
```diff
@@ -113,6 +142,8 @@ exports.playCardAndWord = (username, cardId, word) => {
             cardId: cardId
         };
         cards.push(card);
+        let playerCards =  players.find(player => player.username === currentPlayer.username).cards;    
+        playerCards.find(playedCard => playedCard.id.toString() === cardId).played = true;
```
tillywoodfield, line 138: I don't think this does anything

server/services/gameLogic.js
```diff
@@ -131,6 +162,8 @@ exports.playCard = (username, cardId) => {
             cardId: cardId
         };
         cards.push(card);
+        let playerCards =  players.find(player => player.username === username).cards;    
+        playerCards.find(playedCard => playedCard.id.toString() === cardId).played = true;
```
tillywoodfield, line 158: Same as above

sibela46: Why not?

tillywoodfield: Because you're making a copy of a list and then changing something in it right? Either way, I don't think we need a 'played' attribute, as we have a list of played cards, so `.played` just becomes `cards.includes`

server/services/gameLogic.js
Ro4052: The return from find will be the actual object, so it should work.

server/services/gameLogic.js
Ro4052: Just to clarify, JavaScript is pass by value but this doesn't hold where objects are concerned. Unless you're doing object assign, spread, etc. passing an object is technically pass by reference. So in this case, the `find` will return an object which is a reference to the one in `playerCards`.

sibela46: @tillywoodfield I thought you told me to put a played attribute in the player cards. It's needed because whenever a player refreshes a page, you don't want to return their original hand, you only want to return the ones that they haven't played yet.

cypress/integration/auth_spec.js
```diff
@@ -2,21 +2,22 @@
...
-    cy.visit('http://localhost:8080/dashboard') 
-
-
-    // we should not be redirected to /dashboard
-    cy.url().should('include', '/')
-  }); 
+    it('stops you getting to /dashboard', function () {
+        cy.visit('/dashboard') 
+        // we should not be redirected to /dashboard
+        cy.url().should('include', '/')
```
tillywoodfield, line 8: Does `/dashboard` include `/`?

sibela46: @iatkins

cypress/integration/dashboard_spec.js
```diff
@@ -1,3 +1,40 @@
 describe('The Dashboard Page', function () {
+    beforeEach(() => {
+        cy.request({
```
tillywoodfield: With e2e tests I thought we were supposed to interact with the UI rather than send requests ourselves?

server/controllers/index.js
```diff
@@ -58,6 +63,19 @@ router.get('/api/start', auth, (req, res) => {
...
+/* Check if player is logged in */
+router.get('/api/end', auth, (req, res) => {
+    const user = currentUsers.find((user) => user.username === req.session.user);
+    currentUsers = currentUsers.filter((otherUser) => otherUser !== user);
+    if (user) {
+        gameLogic.endGame();
+        req.session.destroy();
+        res.sendStatus(200);
+    } else {
+        res.status(404).json({message: "Cannot end game: user does not exist."});
```
tillywoodfield: 400?

server/controllers/index.js
```diff
@@ -58,6 +63,19 @@ router.get('/api/start', auth, (req, res) => {
     res.sendStatus(200);
 });
 
+/* Check if player is logged in */
+router.get('/api/end', auth, (req, res) => {
+    const user = currentUsers.find((user) => user.username === req.session.user);
+    currentUsers = currentUsers.filter((otherUser) => otherUser !== user);
```
tillywoodfield: If you're ending the game, shouldn't you be setting `currentUsers = []` rather than just removing the player that made the request?

server/controllers/index.js
```diff
@@ -58,6 +63,19 @@ router.get('/api/start', auth, (req, res) => {
...
+    if (user) {
```
tillywoodfield: Also as we passed `auth` into the request, I think this `if` is redundant?

cypress/integration/dashboard_spec.js
sibela46: Hmm, yeah I guess so, I'll change this now

server/services/cards.js
```diff
@@ -5,15 +5,17 @@ module.exports.assign = (users, number_of_cards) => {
...
+        let availableGlobally = allCards.some(card => card.id !== random);
+        let inCurrentHand = cardsSet.some(card => card.id === random);
+        
+        while (availableGlobally || inCurrentHand) {            
+            random = Math.floor(Math.random() * (max - min) + min);
+            availableGlobally = allCards.some(card => card.id === random);
+            inCurrentHand = cardsSet.some(card => card.id === random);
         }
-        cardsSet.push(random);
+        cardsSet.push({ id: random, played: false });
```
tillywoodfield, line 18: Are you using this `played` attribute on the front end? Because I don't think we need it on the back end, as we keep a list of played cards

server/services/gameLogic.js
tillywoodfield: Ah okay I didn't know that, thanks! I still think we don't need this in the first place though

server/services/socket.js
```diff
@@ -57,36 +79,39 @@ exports.emitWord = word => {
...
-exports.promptPlayersVote = currentPlayer => {
+const promptPlayersVote = currentPlayer => {
     // console.log("promptPlayersVotes");
     for (let socket of sockets) {
         if (socket.handshake.session.user !== currentPlayer.username) {
             socket.emit("vote");
         }
     }
 }
+exports.promptPlayersVote = promptPlayersVote
```
tillywoodfield: Linting error, no new line at end of file

tillywoodfield (approved):

PR #88, opened 2019-07-10
Enabled /api/end request only in dev mode

sibela46:


server/controllers/index.js
```diff
@@ -63,13 +63,15 @@ router.get('/api/start', auth, (req, res) => {
...
 });
 
-/* Check if player is logged in */
-router.get('/api/end', auth, (req, res) => {
-    currentUsers = []
-    gameLogic.endGame();
-    req.session.destroy();
-    res.sendStatus(200);
-});
+/* Check if in dev mode, and enable end game request */
```
tillywoodfield, line 66: I think there are cases where NODE_ENV can be different things e.g. development, production, etc. so possibly change this to `if (process.env.NODE_ENV === 'testing')` and then set this environment variable when we start the server by doing `npx cross-env NODE_ENV='testing' nodemon server.js`? Also could the api call be more descriptive like 'reset server' or something?

scottlogic-alex: agreed -- the content is important, not just the truthiness. Tilly's code looks correct.

sibela46: Thanks! Adding it now...

server/controllers/index.js
```diff
@@ -63,13 +63,15 @@ router.get('/api/start', auth, (req, res) => {
...
+if (!process.env.NODE_ENV) {
+    router.get('/api/end', (req, res) => {
```
scottlogic-alex: HTTP specification says that GET requests are not supposed to have side-effects.
I know GET is more convenient (because browser address bar gives you a cheeky shortcut to send GET requests).

you could do it as a POST with no payload.

server/controllers/index.js
```diff
@@ -63,13 +63,15 @@ router.get('/api/start', auth, (req, res) => {
...
-/* Check if player is logged in */
-router.get('/api/end', auth, (req, res) => {
-    currentUsers = []
-    gameLogic.endGame();
-    req.session.destroy();
-    res.sendStatus(200);
-});
+/* Check if in dev mode, and enable end game request */
+if (process.env.NODE_ENV === 'testing') {
+    router.get('/api/reset-server', (req, res) => {
```
scottlogic-alex: HTTP specification says that GET requests are not supposed to have side-effects.
I know GET is more convenient (because browser address bar gives you a cheeky shortcut to send GET requests).

you could do it as a POST with no payload.

scottlogic-alex (approved):

PR #91, opened 2019-07-10
Alsatian/e2e played cards

sibela46:


server/services/auth.js
```diff
@@ -1,4 +1,7 @@
 module.exports = (req, res, next) => {
+    if (process.env.NODE_ENV === 'testing' && !req.session.user) {
+        req.session.user = 'halfling'
+    }
     if (req.session.user) next();
     else res.sendStatus(401);
 }
```
tillywoodfield: lint eof

server/services/gameLogic.js
```diff
@@ -106,7 +106,7 @@ exports.endGame = () => {
     cards = []
     currentPlayer = null
     currentWord = ''
-    roundNum = 0;
+    roundNum = 0
```
tillywoodfield: For consistency can we have semicolons

cypress/integration/played_cards_spec.js
```diff
@@ -0,0 +1,37 @@
...
+        });
+        cy.visit('/')
+    });
+
+    const cardsNumber = 3;
+    const numberOfPlayers = 2;
+
+    /* GAME STARTS AND PLAYERS ARE GIVEN CARDS */
+    describe("on second player's turn, after first player has played", () => {
+        it('player is prompted to pick a card and all the cards are displayed on the end of his turn', () => {
```
Ro4052: Surprised you went for "his" 😄 
```suggestion
        it('player is prompted to pick a card and all the cards are displayed on the end of their turn', () => {
```

package.json
```diff
@@ -20,19 +20,18 @@
...
-    "start:devserver": "npx cross-env NODE_ENV=testing nodemon server.js --watch ./server --watch ./server.js",
+    "start:devserver": "concurrently \"nodemon ./server/proxy/server.js\" \"npx cross-env NODE_ENV=testing nodemon --inspect server.js --watch ./server --watch ./server.js\"",
     "start:devclient": "react-scripts start",
     "build": "react-scripts build",
     "test:client": "react-scripts test --coverage --watchAll=false",
     "eject": "react-scripts eject",
     "cy:run": "cypress run",
     "cypress:open": "cypress open",
-    "test:e2e": "concurrently --success first --kill-others \"npx cross-env NODE_ENV=testing node server.js\" \"node run-cypress-test.js -r tap\""
+    "test:e2e": "concurrently --success first --kill-others \"nodemon ./server/proxy/server.js\" \"npx cross-env NODE_ENV=testing node server.js\" \"node run-cypress-test.js -r tap\""
```
Ro4052: Probably don't need to used `nodemon` here, you don't need the restart functionailty do you? Also, why `npx`? Should we not just add the package we're using to the dev dependencies and use `npm`?

server/proxy/server.js
```diff
@@ -0,0 +1,61 @@
...
+    const instance = axios.create({
+        baseURL: url,
+        timeout: 500,
+    });
+    
+    socket.on('connect', () => {
+        instance.post('auth/login', { 
+            username: "halfling" 
+        })
+        .then(() => {
```
Ro4052: Can be a one-liner. Not an issue, just think `return` is kind of ugly in this context.

package.json
```diff
@@ -20,19 +20,18 @@
     "redux": "^4.0.1",
     "redux-thunk": "^2.3.0",
     "socket.io": "^2.2.0",
-    "socket.io-client": "^2.2.0",
     "socket.io-mock": "^1.2.3"
   },
   "scripts": {
     "start": "node server.js",
-    "start:devserver": "npx cross-env NODE_ENV=testing nodemon server.js --watch ./server --watch ./server.js",
+    "start:devserver": "concurrently \"nodemon ./server/proxy/server.js\" \"npx cross-env NODE_ENV=testing nodemon --inspect server.js --watch ./server --watch ./server.js\"",
```
Ro4052, line 28: What is this new script doing, you're running two servers? By the name it has something to do with proxying?

server/proxy/server.js
sibela46: @scottlogic-alex suggested using return to chain them all together and have the same catch block I think

package.json
sibela46: Yeah, it's supposed to simulate a second client and makes requests to the original server when prompted to from the first client haha, it's a bit confusing, I'll explain at the stand up

server/proxy/server.js
scottlogic-alex: Ronan's not talking about the topology of the Promise chain.
he's saying your arrow function could be simplifed

Ro4052: If you don't have curly braces, it's an implicit `return`.

package.json
sibela46: That's a good point, I didn't pay attention to this.. I'll change it now

server/proxy/server.js
scottlogic-alex:
```js
.then(() => instance.get('api/cards'))
```
or if you're an absolute madman you could get rid of the arrow function entirely (don't do this):
```js
.then(instance.get.bind(null, 'api/cards'))
```

package.json
Ro4052: OK, that's fine 😃

server/proxy/server.js
```diff
@@ -0,0 +1,61 @@
...
+            return instance.get('api/cards')
+        })
+        .then((response) => {
+            cards = response.data;
```
scottlogic-alex: assigning `cards` to a variable in a parent closure widens its lifecycle.

if all you wanted to do was pass it to the next promise in the chain, there are better patterns for this

EllieHield (dismissed):
Looks good! 
edit - I didn't spot the other guys queries so resolve those before merging

server/proxy/server.js
scottlogic-alex: here's one (of many) ways:
```js
.then(response => Promise.all([
	response.data,
	instance.get('api/start')
	]))
.then(([cards, startResponse]) => {
    if (startResponse.status === 200) {
        return instance.post('api/playCardWord', {
            card: cards[0].id.toString(),
            word: "Hello"
        })
    }
}) 
```

sibela46: Yeah, that's all I was using it for. Thanks, that feels a lot cleaner!

scottlogic-alex: or if a Promise pyramid is acceptable to you, you can use the Promise's closure:
```js
.then(response => {
    cards = response.data;
    return instance.get('api/start')
    .then(startResponse => {
	    if (startResponse.status === 200) {
	        return instance.post('api/playCardWord', {
	            card: cards[0].id.toString(),
	            word: "Hello"
	        })
	    }
	}) 
})
```
but this makes it a bit more complicated to reason about `catch()` behaviour

server/services/gameLogic.js
scottlogic-alex: is there no linter or editorconfig yet?

server/services/auth.js
```diff
@@ -1,4 +1,7 @@
 module.exports = (req, res, next) => {
+    if (process.env.NODE_ENV === 'testing' && !req.session.user) {
+        req.session.user = 'halfling'
```
scottlogic-alex: does this mean that all API accesses initiated by tests will now be authorised? how will you test the authorisation logic effectively?

package.json
scottlogic-alex: > Should we not just add the package we're using to the dev dependencies and use npm?

that's more steps, right? and it relies on node_modules' being pathed correctly.

tillywoodfield: `npx` is for the `cross-env` thing, it's needed for adding the environment variables that we want apparently, Alex and I found it

server/proxy/server.js
```diff
@@ -0,0 +1,61 @@
+const client = require('socket.io-client')
```
tillywoodfield: This whole file has half semicolons and half not semicolons

server/services/auth.js
sibela46: I did not think of this... let me see if the session is set without this line

tillywoodfield: It's not the api calls initiated by tests, it's the server initiated with the testing environment variable, right? I may have misunderstood your question

package.json
scottlogic-alex: cypress is stuck inside a web browser, so all requests it makes will use the same cookie store. this means cypress cannot send requests on behalf of multiple users.  
cypress needs a way to escape the web browser, and control an HTTP client that has a _separate_ cookie store.  
the only way to escape the web platform is via HTTP (we can't run a shell script).  
so, we arrange for a server to be running on 8081. it's a slave that will run any HTTP request you ask.  
and since it's a server (listens for requests from cypress) and also a client (sends HTTP requests to our backend): we call it a proxy. it proxies requests from cypress to the backend.

package.json
Ro4052: OK, maybe there's more to `npx` than I thought.

cypress/integration/home_page_spec.js
```diff
@@ -6,29 +6,31 @@ describe('The Home Page', function() {
...
+            url: '/api/reset-server',
+            method: 'GET'
+        });
+        cy.visit('/');
     })
+
     it('sets auth cookie when logging in via form submission', function () {
-    // destructuring assignment of the this.currentUser object
-    const username = 'jane'
+        // destructuring assignment of the this.currentUser object
```
scottlogic-alex: there's no destructuring assignment here :P

sibela46: this is from another pr and is already on develop, I just merged develop into mine

tillywoodfield (approved):

PR #96, opened 2019-07-11
Player is able to vote for other players' card only

sibela46:


src/store/playerReducer.js
```diff
@@ -14,7 +16,7 @@ export const initialState = {
 
 const cardReducer = (state = initialState, action) => {
     switch (action.type) {
-        case types.RESET_PLAYER_STATE:
+        case types.RESET_PLAYER_STATE:
```
tillywoodfield: Extra whitespace

server/services/gameLogic.js
```diff
@@ -185,4 +185,4 @@ const playerHasVoted = (username) => votes.filter(vote => vote.username === user
 const allPlayersPlayedCard = () => players.every(player => playerHasPlayedCard(player.username));
 
 /* Returns true if all players (apart from the current player) have voted this round */
-const allPlayersVoted = () => players.filter(player => player.username !== currentPlayer.username).every(player => playerHasVoted(player));
+const allPlayersVoted = () => players.filter(player => player.username !== currentPlayer.username).every(player => playerHasVoted(player.username));
```
Ro4052:
```suggestion
const allPlayersVoted = () => players.every(player => player.username === currentPlayer.username || playerHasVoted(player.username));
```
You don't need to loop twice this way. However, it won't have much impact so if you want to leave the `filter` in for clarity, go for it.

src/components/Auth/Auth.jsx
```diff
@@ -45,8 +45,8 @@ export class Auth extends React.Component {
 }
 
 const mapDispatchToProps = (dispatch) => ({
-    resetState: (id) => dispatch(resetState),
-    resetPlayerState: (id) => dispatch(resetPlayerState)
+    resetState: () => dispatch(resetState()),
+    resetPlayerState: () => dispatch(resetPlayerState())
```
Ro4052, line 49: How was this working before?

src/components/Cards/PlayedCards/PlayedCards.jsx
```diff
@@ -1,20 +1,45 @@
...
+
+    getCardClass(id) {
+        return (this.props.voteCard && !this.props.votedCard && (id !== this.props.playedCard.toString())) ? styles.allCards : styles.allCardsDisabled;
+    }
+
     render() {
         return (
             <ul id="played-cards">
-                {this.props.cards.map(card => <img id={"card-" + card.cardId} key={card.cardId} className={styles.allCards} alt={"card-" + card.cardId} src={require(`../cards/card (${card.cardId}).jpg`)}/>)}
+                {this.props.cards.map(card => <img id={"card-" + card.cardId} key={card.cardId} className={this.getCardClass(card.cardId)} alt={"card-" + card.cardId} src={require(`../cards/card (${card.cardId}).jpg`)} onClick={this.voteForCard.bind(this)}/>)}
```
Ro4052: You've bound `this.voteForCard` twice.

src/components/Cards/PlayedCards/PlayedCards.jsx
```diff
@@ -1,20 +1,45 @@
...
             </ul>
         );
     }
 }
 
-const mapStateToProps = (state) => ({
-    cards: state.reducer.currentCards
+const mapStateToProps = (state) => {
+    return ({
+        cards: state.reducer.currentCards,
```
Ro4052: I'd consider not having a reducer called `reducer`. I don't think it's very descriptive.

src/components/PlayerInfo/VoteCard/VoteCard.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import styles from '../PlayWordAndCard/PlayWordAndCard.module.css';
+
+export class VoteCard extends React.PureComponent {
```
Ro4052: This component doesn't have any state. If you want to optimise, you should make this a function component, rather than a Pure component.

src/store/playerActions.js
```diff
@@ -59,11 +65,43 @@ export const playWord = (word) => ({
...
+
+export const voteForCardFailure = (error) => ({
+    type: types.VOTE_FOR_CARD_FAILURE,
+    payload: { error }
+});
+
+export const voteForCard = (id) => {
+    return (dispatch) => {
+        dispatch(voteForCardBegin());
+        return axios.post('/api/voteCard', {
```
Ro4052: Two points here:
- Do you need to return the `axios` request?
- If you want to clean up thunks, you can do something like this
```
export const voteForCard = id => dispatch => {
  // Function
}
```

server/services/gameLogic.js
```diff
@@ -53,7 +53,7 @@ exports.getCurrentPlayer = () => currentPlayer;
 exports.getCardsByUsername = (username) => players.find(player => player.username === username).cards.filter(card => card.played === false);
 
 /* Returns true if the player is able to join the game, false otherwise */
-exports.canJoinGame = (username) => (status === statusTypes.NOT_STARTED && !players.some(player => player.username === username));
+exports.canJoinGame = (username) => ((status === statusTypes.NOT_STARTED || status === statusTypes.GAME_OVER) && !players.some(player => player.username === username));
```
tillywoodfield: When I added the `statusTypes.GAME_OVER` I was thinking it would be used for the stage in the game when scores are displayed etc so we don't want people joining the game at that point. Then when someone then gets a new game ready the status would go back to `NOT_STARTED` and people could join again

sibela46: Right, I see... I'll remove that, it was mainly for testing if the game finished when all the players had voted

src/components/Auth/Auth.jsx
sibela46: It wasn't :D

src/components/Cards/PlayedCards/PlayedCards.jsx
sibela46: I think this one was needed for accessing the event in the function? Or is the top one sufficient?

src/services/socket.js
```diff
@@ -1,7 +1,7 @@
 import io from 'socket.io-client';
 import { dispatch } from '../store/store';
 import { setCurrentWord, setStatus, setRoundNumber, setCurrentPlayer, setPlayers, setCurrentCards, setSocket } from '../store/actions';
-import { setPlayWordAndCard, setPlayCard, setPlayedCard } from '../store/playerActions';
+import { setPlayWordAndCard, setPlayCard, setPlayedCard, setVoteCard, setVotedCard, playWord, resetFinishRound } from '../store/playerActions';
```
tillywoodfield, line 4: Forgot to change this but you can change this to `import { types } from '../store/playerActions` and then use `types.THE_ACTION_TYPE` everywhere

src/components/Auth/Auth.jsx
tillywoodfield: My bad 😆

server/services/gameLogic.js
```diff
@@ -185,4 +185,4 @@ const playerHasVoted = (username) => votes.filter(vote => vote.username === user
 const allPlayersPlayedCard = () => players.every(player => playerHasPlayedCard(player.username));
 
 /* Returns true if all players (apart from the current player) have voted this round */
-const allPlayersVoted = () => players.filter(player => player.username !== currentPlayer.username).every(player => playerHasVoted(player.username));
+const allPlayersVoted = () => players.every(player => player.username === currentPlayer.username || playerHasVoted(player.username));
```
tillywoodfield, line 188: I think the way it was before was more readable, how come you changed it?

sibela46: Ronan said there's no need of looping through the list twice

sibela46: I could change it back to what it was before - pick readability or efficiency? :D I don't mind in this situation

src/components/Cards/PlayedCards/PlayedCards.jsx
Ro4052: `bind` is only needed to access the `this` of the class instance. You only need to bind it once, the one in the constructor is enough.

server/services/gameLogic.js
Ro4052: Yeah, that's the choice. I just wanted to highlight another option.

src/services/socket.js
sibela46: I think cause we're importing the actual functions, not the types, they need to stay separate?

src/components/Cards/PlayedCards/PlayedCards.jsx
sibela46: Okay, awesome. Thanks!

src/services/socket.js
tillywoodfield: Oh yeah I forgot these weren't the action types

tillywoodfield (approved):

PR #98, opened 2019-07-12
Added e2e tests for voting

sibela46:


cypress/integration/play_word_and_card_spec.js
```diff
@@ -1,6 +1,7 @@
 describe('Play word and card', () => {
     beforeEach(() => {
         cy.login('username');
+        cy.wait(500);
```
tillywoodfield, line 4: Do you want to move this into the startGame function itself so we don't have to repeat the wait everywhere?

cypress/integration/vote_for_card_spec.js
```diff
@@ -0,0 +1,42 @@
...
+                cy.get('[data-cy="type-word"]').type('word');
+                cy.get('[data-cy="send-word"]').click();
+                cy.get('[data-cy="my-cards"] > img').first().click();
+            } else {
+                cy.get('[data-cy="my-cards"] > img').first().click(); 
+            }
+        });
+    });
+
+    describe('on send a card and vote for one', () => {
```
tillywoodfield: I don't think these `describe`s and `it`s are super descriptive

cypress/integration/vote_for_card_spec.js
```diff
@@ -0,0 +1,42 @@
+describe('Vote for a card', () => {
+    beforeEach(() => {
+        cy.login('unicorn');
+        /* Connect a second user using our fake client */
+        cy
+            .request(`http://localhost:12346/connect?url=${encodeURIComponent(Cypress.config().baseUrl)}`)
```
scottlogic-alex, line 6: maybe use a variable so that `http://localhost:12346` needn't be hardcoded and copy-pasted

scottlogic-alex: worth looking at the `URL` class to safely construct URLs programmatically (i.e. without relying on string concatenation). this may also provide a better alternative to URI-encoding query params on a case-by-case basis

cypress/integration/vote_for_card_spec.js
```diff
@@ -0,0 +1,58 @@
...
+    describe('on player plays a card and then votes for one', () => {
+        it('displays all played cards after current player plays their card', () => {
+            cy.get('[data-cy="played-cards"]').children().its('length').should('eq', 2);
+        });
+        it('player is prompted to vote for a card', () => {
+            cy.get('[data-cy="vote-card"]');
+        });
+        it("player is able to vote for the other player's card only", () => {
+            cy.get('[data-cy="played-cards"] > img').first().then(($img) => {
+                if ($img.hasClass('Cards_allCards__2SdBy')) {
```
Ro4052, line 33: Just a warning about this kind of thing. These class names can change because they depend on file location (among other things). I'm not saying you shouldn't do this (it's definitely the easiest option) but you may have to update in the future.

src/components/Cards/PlayedCards/PlayedCards.jsx
```diff
@@ -31,7 +31,7 @@ export class PlayedCards extends React.Component {
             <ul id="played-cards" styles={styles.playedCards} >
                 {this.props.cards.map(card => 
                     <li className={styles.playedCard} key={card.cardId}>
-                        {(this.props.allVotes) && (this.props.status === 'DISPLAY_ALL_VOTES') ? <span>{this.getVotesForCard(card.cardId)}</span> : ""}
+                        {(this.props.allVotes && this.props.status === 'DISPLAY_ALL_VOTES') ? <span>{this.getVotesForCard(card.cardId)}</span> : ""}
```
Ro4052:
```suggestion
                        {(this.props.allVotes && this.props.status === 'DISPLAY_ALL_VOTES') && <span>{this.getVotesForCard(card.cardId)}</span>}
```

cypress/integration/vote_for_card_spec.js
sibela46: Yeah, unfortunately there's literally no way you can check if something is a substring of a className in cypress, we'll have to figure out another way in the future :(

cypress/integration/vote_for_card_spec.js
sibela46: I tried switching it, but it didn't work straight away, so I'll create an issue and put more thought into it

tillywoodfield (approved):

PR #121, opened 2019-07-15
Reconnecting socket on refresh

sibela46:


src/components/Login/Login.jsx
```diff
@@ -76,7 +76,7 @@ export class Login extends React.Component {
                     </div>
                     <Monster />
                     <form className={styles.loginForm} onSubmit={this.sendLogin.bind(this)}>
-                        <h3 data-cy="login-error" className={styles.errorText}>{}</h3>
+                        <h3 data-cy="login-error" className={styles.errorText}>{(typeof this.props.error === 'string') ? this.props.error : ""}</h3>
```
tillywoodfield, line 79: Does this need to be a ternary?

sibela46: There seems to be a point where I'm setting this error to be the actual error we receive from the HTTP request, which is not a string, so React gives an error when rendering it in an h3. I was thinking that when we do the error handling ticket, this can all be fixed

EllieHield, line 79: out of curiosity what other types might the error be?

src/store/playerActions.js
```diff
@@ -148,21 +148,22 @@ export const authenticateUser = () => (dispatch) => {
...
             dispatch(authSuccess(res.data.cookie));
         } else if (res.status === 401) {
-            dispatch(authFailure());
+            dispatch(authFailure("Unauthorised"));
             dispatch(resetState());
             dispatch(resetPlayerState());
         } else {
             const error = new Error(res.error);
-            dispatch(authFailure());
+            dispatch(authFailure(""));
```
EllieHield, line 159: What does this do? Is it worth dispatching this with an empty string?

EllieHield (commented):
Looks good, just a couple of questions but not necessarily anything wrong

src/components/Login/Login.jsx
EllieHield: Ok, I think that answers my question as well, it would be good to remove the type checking at some point but if that needs to be in the future then that's fine

sibela46: I think it's an object when it's produced from the axios request

src/store/playerActions.js
sibela46: No, but again it's left for whoever is doing the proper error handling (we haven't yet figure out how to deal with the situation in this else case)

src/components/Login/Login.jsx
tillywoodfield: Can't it just start as an empty string and then set the string ourselves when we receive an error from an axios request?

EllieHield (approved):
As long as there are tickets for those two items that are still to-do I'm happy for it to go in :)

PR #122, opened 2019-07-16
Beagle/profanity

iatkins:


server/controllers/index.js
```diff
@@ -78,7 +79,13 @@ if (process.env.NODE_ENV === 'testing') {
 /* Current player plays a card and a word */
 router.post('/api/playCardWord', auth, (req, res) => {
     gameLogic.playCardAndWord(req.session.user, req.body.card, req.body.word);
-    res.sendStatus(200);
+    
+    if (!profanities.notProfanity(req.body.word)) {
+        res.sendStatus(200);
+        console.log("Profane");
```
Ro4052: I might be wrong but wouldn't "profane" mean there _is_ profanity?

server/services/profanityCheck.js
```diff
@@ -0,0 +1,15 @@
+const fs = require('fs');
+const wordListPath = require('word-list');
+const profanities = fs.readFileSync(wordListPath, 'utf8').split('\n');
+
+var notProfanity = function(text){
```
Ro4052: Can avoid using the `var` keyword?

server/services/profanityCheck.js
```diff
@@ -0,0 +1,15 @@
...
+    var returnVal = false; 
+    for (var i = 0; i < profanities.length; i++) {
+        if(text.toLowerCase().includes(profanities[i].toLowerCase())){
+            returnVal = true;
+            break;
+        }
+    }
+    return returnVal;
```
Ro4052: This looks fine, even if I don't love `for` loops in JS! A small change you could do though, is to get rid of `returnVal`. Just return `true` in the `if` statement and return `false` at the end of the function.

src/components/Login/Login.jsx
```diff
@@ -103,11 +103,14 @@ export class Login extends React.Component {
...
-                        <span className={styles.letter} data-letter="i">i</span>
-                        <span className={styles.letter} data-letter="t">t</span>
+                        <span className={styles.letter} data-letter="T">T</span>
+                        <span className={styles.letter} data-letter="e">e</span>
+                        <span className={styles.letter} data-letter="l">l</span>
+                        <span className={styles.letter} data-letter="l">l</span>
+                        <span className={styles.letter} data-letter="T">T</span>
+                        <span className={styles.letter} data-letter="a">a</span>
+                        <span className={styles.letter} data-letter="l">l</span>
+                        <span className={styles.letter} data-letter="e">e</span>
```
Ro4052: If you're trying to avoid Copyright, Telltale is a pretty big [games brand](https://telltale.com/).

Ro4052: Oh, this might also look nicer if you didn't write out each `span` but had a map instead.

server/services/profanityCheck.js
Ro4052: [Wikipedia](https://en.wikipedia.org/wiki/Profanity) seems to agree.

server/controllers/index.js
Ro4052: [Wikipedia](https://en.wikipedia.org/wiki/Profanity) seems to agree.

iatkins: yep! have since changed this, it's because at first I thought word-list contained bad words but it infact contains good words

server/services/profanityCheck.js
iatkins: have turned it into a one liner

server/controllers/index.js
Ro4052: Oh, I see : )

src/components/Login/Login.jsx
iatkins: that's a shame, will look at other names

server/services/profanityCheck.js
```diff
@@ -0,0 +1,6 @@
+const fs = require('fs');
+const wordListPath = require('word-list');
+const validWords = fs.readFileSync(wordListPath, 'utf8').split('\n');
+
+const isValidWord = word => validWords.some(validWord => validWord.toLowerCase() === word.toLowerCase());
```
jburton-scottlogic: would it be worth making the validWords file all lowercase and saving computation time? For every word that's being checked your going to have to lowercase every word in your dictionary

iatkins: good idea, ill add that!

src/components/Login/Login.jsx
iatkins: done the map, changed the name to 'Telltales' ... hoping that 's' makes all the difference?

server/services/gameLogic.js
```diff
@@ -12,7 +12,7 @@ const statusTypes = {
 };
 
 const rounds = 3;
-const minPlayers = 3;
+const minPlayers = 0;
```
Ro4052: Is this just for testing? I thought you needed 3 players.

src/components/Login/Login.jsx
```diff
@@ -63,16 +63,12 @@ export class Login extends React.Component {
...
                 <div className={styles.loginPage}>
-                     <div className={styles.foo}>
-                        <span className={styles.letter} data-letter="D">D</span>
-                        <span className={styles.letter} data-letter="i">i</span>
-                        <span className={styles.letter} data-letter="X">X</span>
-                        <span className={styles.letter} data-letter="i">i</span>
-                        <span className={styles.letter} data-letter="t">t</span>
-
+                    <div className={styles.foo}>
+                        {letters.map((letter, key) => <span key={key} className={styles.letter} data-letter={letter}>{letter}</span>)}
```
Ro4052, line 76: I'm not saying you should change this (hopefully you won't need to change the name again) but I was thinking that you'd have a string rather than an array of letters. Then you'd split the string before mapping.

src/components/PlayWord/PlayWord.jsx
```diff
@@ -19,7 +20,15 @@ export class PlayWord extends React.Component {
     }
 
     sendMessage() {
-        this.props.playWord(this.state.currentValue);
+        axios.post('/api/validWord', {
+            word: this.state.currentValue
+        })
+        .then(() => {
+            this.props.playWord(this.state.currentValue);
```
Ro4052, line 28: So I'm thing this should move into the actions. You'd call an action with `this.state.currentValue` and update the state from there.

server/services/gameLogic.js
iatkins: yep just for testing

src/components/Login/Login.jsx
iatkins: oh I see, that makes sense!

PR #123, opened 2019-07-16
Filled the else cases for error handling

sibela46:


server/services/gameLogic.js
```diff
@@ -73,7 +73,7 @@ exports.quitGame = player => {
         socket.emitPlayers(getPlayers());
         return true;
     } else {
-        // Game has started, player can't quit
+        // Game has started, player can't quit, server is responsible for generating the error
         return false;
     }
```
Ro4052, line 77: There are a few place where you're just returning in the `else` and you don't need the `else`.

src/components/Login/Login.jsx
```diff
@@ -62,7 +62,7 @@ export class Login extends React.Component {
         return (this.props.error) ? {border: '2px solid #EA3546'} : {};
     }
 
-    render() {
+    render() {
```
Ro4052: Blank spaces.

src/store/playerActions.js
```diff
@@ -116,18 +136,18 @@ export const resetCookieSuccess = () => ({
...
     })
-    .then(() => {
-        connectSocket();
-        dispatch(logInSuccess(username))
+    .then((res) => {
+        if (res.status === 200) {
+            connectSocket();
+            dispatch(logInSuccess(username))
+        } else {
+            dispatch(logInFailure(res.data.message));
```
Ro4052, line 147: Do you need the `else` _and_ the `catch`.

Ro4052 (commented):
This was just a quick look.

src/store/playerActions.js
sibela46: Yeah, if the error's status is different from the ones I'm validating.. for example, when it's an internal server error 500

server/controllers/index.js
```diff
@@ -8,75 +8,86 @@ let currentUsers = [];
...
 });
 
 /* Log out the user */
-router.post('/auth/logout', (req, res) => {
+router.get('/auth/logout', (req, res) => {
     const user = currentUsers.find((user) => user.username === req.session.user);
-    if (user) {
-        currentUsers = currentUsers.filter((otherUser) => otherUser !== user);
-        if (gameLogic.quitGame(user)) {
+    if (user) { /* User making the request exists */
```
tillywoodfield: Don't need to check if user here, you can just use the auth service file

sibela46: true

server/controllers/index.js
```diff
@@ -8,75 +8,86 @@ let currentUsers = [];
...
         res.status(404).json({message: "Cannot log out: user does not exist."});
     }
 });
 
 /* Get the players list of cards */
 router.get('/api/cards', auth, (req, res) => {
     const cards = gameLogic.getUnplayedCardsByUsername(req.session.user);
-    res.status(200).json(cards);
+    if (cards) res.status(200).json(cards);
+    else res.status(400).json({message: "Cannot find cards: user does not exist."});
```
tillywoodfield: 404 for user not existing?

server/controllers/index.js
```diff
@@ -8,75 +8,86 @@ let currentUsers = [];
...
 
 /* Resest the cookie and destroy the session */
 router.get('/api/reset-cookie', auth, (req, res) => {
-    req.session.destroy();
-    res.sendStatus(200);
+    if (req.session) {
+        req.session.destroy();
+        res.sendStatus(200);
+    } else {
+        res.status(400).json({message: "Cannot destroy session as it does not exist"});
```
tillywoodfield: Another 404 for not existing

tillywoodfield:
I don't really like the whole returning true or false from gameLogic functions, as it isn't very readable, is there a way to throw an error in the gameLogic function and catch it in the controller to then return an error response?

sibela46:
> I don't really like the whole returning true or false from gameLogic functions, as it isn't very readable, is there a way to throw an error in the gameLogic function and catch it in the controller to then return an error response?

@Ro4052 is there?

Ro4052:
> > I don't really like the whole returning true or false from gameLogic functions, as it isn't very readable, is there a way to throw an error in the gameLogic function and catch it in the controller to then return an error response?
> 
> @Ro4052 is there?

Have you looked into `try` and `catch` blocks?

server/controllers/index.js
```diff
@@ -8,75 +8,81 @@ let currentUsers = [];
 
 /* Check if player is logged in */
 router.get('/auth', (req, res) => {
-    if (req.session.user) {
+    if (req.session.user) { /* The user session exists, authorised */
```
jburton-scottlogic: could use destructuring here
`
const { user } = req.session;
`
then use the user var line 11 and 12.

server/controllers/index.js
```diff
@@ -8,75 +8,81 @@ let currentUsers = [];
...
+        res.status(401).json({ message: "You need to log in with a username to enter the game."});
     }
 });
 
 /* Log in the user */
 router.post('/auth/login', (req, res) => {
     const user = currentUsers.find((user) => user.username === req.body.username);
-    if (!user) {
+    if (!user) { /* User does not already exist */
         req.session.user = req.body.username;
```
jburton-scottlogic, line 24: same here, you can destructure the username from req.body.. also means when building your 'user' object you can do:
`
const { username } = req.body;
const user = {
username
};
`

you dont need to do username: username

server/controllers/index.js
```diff
@@ -8,75 +8,81 @@ let currentUsers = [];
...
-        } else {
-            res.sendStatus(400)
-        }
-    } else {
-        res.status(404).json({message: "Cannot log out: user does not exist."});
+        req.session.destroy();
+        gameLogic.removePlayer(user.username);
+        res.sendStatus(200);
+    } else { /* Game has started, method not allowed */
+        res.status(400).json({message: "Cannot log out of a running game."});
```
jburton-scottlogic: general point around this file / api - if the gameLogic.quitGame() or any others throw an error, are they handled gracefully in the gameLogic method? If not it might be worth using try catch blocks because one error will crash your server for everyone

server/controllers/index.js
sibela46: What does this do exactly? Does it match the field in the object on the right (in our case user) and assign it to our const?

server/controllers/index.js
```diff
@@ -8,75 +8,81 @@ let currentUsers = [];
...
     }
 });
 
 /* Get the players list of cards */
 router.get('/api/cards', auth, (req, res) => {
     const cards = gameLogic.getUnplayedCardsByUsername(req.session.user);
-    res.status(200).json(cards);
+    if (cards) res.status(200).json(cards);
```
jburton-scottlogic: tempting to not use curly braces but i dont think it makes it more readable / cleaner, imo makes it harder but its just an opinion :), same with the else.

also as ronan mentioned in one of the comments, for this, it might be more readable to do soemthign like:
```

if (cards) {
 return res.status(200).json(cards);
}
return res.status(404).json({message: "Cannot find cards: user does not exist."});

```
eliminates the need for an else - whatever you all decide looks better is fine!

server/services/gameLogic.js
```diff
@@ -72,32 +72,37 @@ exports.quitGame = player => {
...
-    playedCards = [];
-    votes = [];
+    if (status === statusTypes.GAME_OVER) {
+        socket.emitEndGame();
+        status = statusTypes.NOT_STARTED;
+        currentPlayer = null;
+        currentWord = '';
+        roundNum = 0;
+        players = [];
+        playedCards = [];
```
jburton-scottlogic: could possibly group a couple of these up e.g.

cleanPlayerData() {
currentWord = '';
    players = [];	        roundNum = 0;
    playedCards = [];	        players = [];
    votes = [];	        playedCards = [];
}

and reuse the logic in the nextRound() method etc.

server/controllers/index.js
sibela46: Yeah, Ronan suggested using try and catch instead as well, I just pushed a new commit

server/controllers/index.js
jburton-scottlogic: yeah exactly, its called destructuring, you can do it with an array or object
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

sibela46: right, okay :) thanks

tillywoodfield (approved):

server/controllers/index.js
Ro4052: I agree with not having `if` statements without braces.

PR #133, opened 2019-07-17
Adding exit code

sibela46:


cypress/integration/vote_for_card_spec.js
```diff
@@ -29,13 +29,13 @@ describe('Vote for a card', () => {
...
-    describe('on vote for a card', () => {
-        beforeEach(() => {
-            cy.voteCard();
-        });
-        // it("displays the votes", () => {
-        //     cy.get('[data-cy="vote-card"]').should('not.exist');
-        //     cy.get('[data-cy="votes"]').should('exist');
-        // });
-    });
+    // describe('on vote for a card', () => {
```
Ro4052, line 32: I would prefer to delete rather than comment.

run-cypress-test.js
```diff
@@ -3,5 +3,13 @@ const server = require('./server/server');
 
 cypress.run()
 .then((results) => {
-  return server.stop()
-});
+  console.log(results.totalFailed);
+  if (results.totalFailed > 0) {
+    return server.fail();
+  }
+  return server.stop();
```
Ro4052, line 10: Could this then be a one-liner?

`results.totalFailed > 0 ? server.fail() : server.stop()`

That would get rid of the log though.

Ro4052 (commented):
You might want to squash these commits 🙈

run-cypress-test.js
sibela46: I wanted the log for debugging, I can probably change this now

PR #147, opened 2019-07-19
Voting card test to what it was before

sibela46:


iatkins (approved):
top stuff, keep it up

PR #171, opened 2019-07-23
Corgi/redux refactor

tillywoodfield:


server/controllers/index.js
```diff
@@ -21,21 +21,19 @@ router.get('/auth', (req, res) => {
...
-        if (gameLogic.canJoinGame(user.username)) { /* Game has not been started yet */
-            currentUsers.push(user);
-            gameLogic.joinGame(user, () => {
-                res.sendStatus(200);
-            });
-        } else { /* Game has already began */
-            res.status(400).json({message: "Game has already started."});
-        }
-    } else { /* Username is taken, conflict error */
+    const { username } = req.body;
```
Ro4052, line 24: Did you change this to 3 space tabs???

server/services/gameLogic.js
```diff
@@ -59,15 +53,14 @@ const getPlayedCards = () => playedCards.map(card => ({ cardId: card.cardId })).
...
 
 /* Add the player to the game if possible */
-exports.joinGame = (user, callback) => {
-    const cards = cardManager.assign(players, rounds);
-    const player = {
-        username: user.username,
-        cards: cards,
-        score: 0
+exports.joinGame = (username) => {
+    if (status !== statusTypes.NOT_STARTED) throw Error("Game has already started");
```
Ro4052: I think this block would look clearer if you used curley braces.

src/components/EndTurn/EndTurn.jsx
```diff
@@ -14,10 +14,10 @@ export class EndTurn extends React.Component {
 
     endTurn() {
         axios.post('/api/playCardWord', {
-            cardId: this.props.playedCard,
-            word: this.props.myWord
+            cardId: this.props.playedCardId,
+            word: this.props.word
         })
-        .then(() => this.props.finishPlayCard(this.props.playedCard))
+        .then(() => this.props.finishPlayCard(this.props.playedCardId))
```
Ro4052: I still think the API calls shouldn't live in the components. Especially ones like this, the result has no direct effect on the component making the call.

src/components/Login/Login.jsx
```diff
@@ -31,9 +31,7 @@ export class Login extends React.Component {
     checkUsernameAllowed(string) {
         // eslint-disable-next-line
         const allowed = /^[A-Za-z0-9]*$/;
-        if (allowed.test(string)){
-            return true;
-        }
+        if (allowed.test(string)) return true;
         return false;
```
Ro4052: Can we change this bit to:
`return allowed.test(string)`
?
Or something similar?

src/components/shared/Auth/Auth.jsx
```diff
@@ -1,45 +1,26 @@
...
-    playedCard: state.playerReducer.playedCard,
-    myWord: state.playerReducer.myWord
+    username: state.authReducer.username
 });
 
 const mapDispatchToProps = (dispatch) => ({
-    authenticateUser: () => dispatch(authenticateUser()),
-    setPlayWord: (bool) => dispatch(setPlayWord(bool)),
-    setPlayCard: (bool) => dispatch(setPlayCard(bool))
+    authenticateUser: () => dispatch(authenticateUser())
```
Ro4052, line 23: I don't think I like this component.

server/controllers/index.js
tillywoodfield: No?

cypress/integration/my_cards_spec.js
```diff
@@ -10,10 +10,9 @@ describe('The Player Cards Component', () => {
         });
     });
 
-    /* TODO: figure out how to test border */
     describe('player clicks on a card', () => {
         it("its highlighted by being moved upwards on the y axis", () => {
-            cy.playCard();
+            cy.get('[data-cy="my-cards"] [data-cy="card"]').first().click();
```
sibela46: how come you're not using one of the commands?

cypress/integration/play_card_spec.js
```diff
@@ -1,19 +1,12 @@
...
+            cy.request(`http://localhost:12346/connect?url=${encodeURIComponent(Cypress.config().baseUrl)}`)
+            .then(() => {
                 cy.startGame();
-                cy.request({
-                    method: 'GET',
-                    url: `http://localhost:12346/playCardWord?url=${encodeURIComponent(Cypress.config().baseUrl)}`,
-                    followRedirect: false
-                }).then(() => {
+                cy.request(`http://localhost:12346/playCardWord?url=${encodeURIComponent(Cypress.config().baseUrl)}`)
+                .then(() => {
```
sibela46: could be one line

src/components/EndTurn/EndTurn.jsx
tillywoodfield: Yeah I didn't want to change too much in one refactor, should I change it?

cypress/integration/play_word_and_card_spec.js
```diff
@@ -1,6 +1,5 @@
 describe('Play word and card', () => {
     beforeEach(() => cy.login('unicorn'));
-
```
sibela46: do we not want the space to separate the functions

src/components/MyCards/MyCards.test.js
```diff
@@ -33,50 +33,50 @@ describe('on clicking an enabled card', () => {
...
-                <Provider store={store}>
-                    <MyCards fetchCards={jest.fn()} myCards={cards} requestPlayCard={requestPlayCard} playWordAndCard={true} />
-                </Provider>
-            );
-            wrapper.find({ 'data-cy': 'card' }).first().simulate('click');
-            expect(requestPlayCard).toHaveBeenCalled();
-            requestPlayCard.mockRestore();
-        }));
-    });
+    // describe('on playWordAndCard flag true', () => {
```
sibela46: Are we gonna use this test eventually? Why is it commented out?

src/components/shared/Auth/Auth.jsx
tillywoodfield: What would you do instead?

src/components/PlayedCards/PlayedCards.jsx
```diff
@@ -12,11 +12,11 @@ export class PlayedCards extends React.Component {
     }
 
     getVotesForCard(card) {
-        return this.props.allVotes.reduce((sum, vote) => sum + (vote.cardId === card.cardId), 0)
+        return this.props.votes.reduce((sum, vote) => sum + (vote.cardId === card.cardId), 0)
```
sibela46: semicolon

src/components/PlayerInteractions/PlayerInteractions.jsx
```diff
@@ -16,24 +16,24 @@ export class PlayerInteractions extends React.Component {
...
-    playCard: state.playerReducer.playCard,
-    playWord: state.playerReducer.playWord,
-    voteCard: state.playerReducer.voteCard,
-    currentWord: state.gameReducer.currentWord,
-    winner: state.gameReducer.winner,
-    finishedRound: state.playerReducer.finishedRound,
-    myWord: state.playerReducer.myWord,
-    playedCard: state.playerReducer.playedCard
+    status: state.dashboardReducer.status,
+    // roundNum: state.playerReducer.roundNum,
```
sibela46: If we don't need the commented variables, we can just delete them now

src/components/shared/Auth/Auth.jsx
sibela46: @Ro4052 what's a good practice to deal with redirecting on unsuccessful authorisation?

cypress/integration/my_cards_spec.js
tillywoodfield: The command is for actually sending the card, whereas this test is for seeing whether it becomes 'selected' when you click on it

src/components/MyCards/MyCards.test.js
tillywoodfield: This flag hasn't been used for a while and so I removed it in the refactor, the test is redundant now

sibela46 (dismissed):

server/controllers/index.js
Ro4052: Oh, never mind, guess I was (not) seeing things. I'm not sure why it says this line has changed.

src/components/EndTurn/EndTurn.jsx
Ro4052: If it's going to be done at some point, you should probably do it now or it'll likely never be done.

tillywoodfield: Good point

src/components/shared/Auth/Auth.jsx
Ro4052: Essentially, I don't think this is something React should care about, since it has nothing to do with rendering. A couple of options I can think of:

- You have this action dispatched whenever the page loads (maybe in some index file) and redirect via `history.push` if necessary
- You only use this `Auth` component for `Dashboard`, so you could just make the call there and redirect

One thing you can do if you want is have some sort of "loading" component. You could show that whenever the UI is waiting for some action to take place. E.g. you could show that while authenticating the user. That's another point though and would work in conjuction with one of the approaches above.

Note: My personal preference would probably be option 1. That is a preference though and I'd be happy with either.

src/components/MyCards/MyCardsActions.js
```diff
@@ -13,25 +18,33 @@ export const fetchCardsFailure = error => ({
...
-export const resetFinishRound = finishedRound => ({
-    type: types.RESET_FINISH_ROUND,
-    finishedRound
-});
+export const selectCard = cardId => (dispatch, getState) => {
+    const status = getState().dashboardReducer.status;
+    if (status === "WAITING_FOR_CURRENT_PLAYER") {
+        dispatch(selectCardSuccess(cardId));
+    } else if (status === "WAITING_FOR_OTHER_PLAYERS") {
+        instance.post('/api/playCard', { cardId })
```
Ro4052: You may, at some point, want to rename this variable as it's not very descriptive. Also, looks like it's a `let` that could be a `const`.

src/components/PlayerInteractions/PlayerInteractionsActions.js
```diff
@@ -0,0 +1,33 @@
+import { types } from './PlayerInteractionsActionTypes';
+import axios from "axios";
+import { removeCard } from '../MyCards/MyCardsActions';
+
+let instance = axios.create({ validateStatus: status => (status >= 200 && status < 500) });
+
+export const endTurn = () => (dispatch, getState) => {
+    const cardId = getState().myCardsReducer.playedCardId;
+    const word = getState().playWordReducer.word;
```
Ro4052: Rather than calling `getState` twice, would it be better to assign an intermediate value?
```
const state = getState();
const cardId = state.myCardsReducer.playedCardId;
const word = state.playWordReducer.word;
```

Ro4052 (commented):
Looking good, must have taken a while to do all this 😄

src/components/MyCards/MyCardsActions.js
tillywoodfield: Which variable, instance?

Ro4052: Whoops, yes, I was referring to instance.

tillywoodfield: Have changed it to a const, what would you rename it to? I think `instance` was just what they called it in whatever examples were found

Ro4052: Even if you just called it `axiosInstance` I think it would be more readable.

cypress/integration/my_cards_spec.js
```diff
@@ -12,8 +12,9 @@ describe('My Cards', () => {
 
     describe('player clicks on a card', () => {
         it("its highlighted by being moved upwards on the y axis", () => {
-            cy.get('[data-cy="my-cards"] [data-cy="card"]').first().click();
-            cy.get('[data-cy="my-cards"]').children().first().should('have.css', "transform", "matrix(1, 0, 0, 1, 0, -30)");
+            cy.get('[data-cy="my-cards"] [data-cy="card"]').first().click().then(() => {
```
Ro4052, line 15: You might be able to use async/await but if this approach works, there's nothing wrong with it.

tillywoodfield: I saw it done like this on the cypress website

Ro4052: To elaborate, you'd need to make the function `async` and do something like:
```
it("its highlighted by being moved upwards on the y axis", async () => {
    await cy.get('[data-cy="my-cards"] [data-cy="card"]').first().click();
    cy.get('[data-cy="my-cards"]').children().first().should('have.css', "transform", "matrix(1, 0, 0, 1, 0, -30)");
```
I'm not sure if you'd need to use `done` somewhere in here.

tillywoodfield: Oh I see, okay

Ro4052: I just realised I hadn't seen your reply when I was writing mine.

sibela46 (approved):

PR #186, opened 2019-07-25
Players redirected to lobby after end of game

sibela46:


server/proxy/server.js
```diff
@@ -143,6 +143,22 @@ app.get('/playCard', (req, res) => {
     .catch((err) =>  res.send(err));
 });
 
+app.get('/voteCard', (req, res) => {
```
jburton-scottlogic, line 146: Struggling to understand what this is doing? Is there a danger of the user being able to pass in their own url here that theyre hosting? or would that even matter?

server/services/socket.js
```diff
@@ -91,6 +91,13 @@ exports.emitWinner = (roomId, player) => io.to(`room-${roomId}`).emit("winner",
 // When game is over, tell the users
 exports.emitEndGame = (roomId) => io.to(`room-${roomId}`).emit("end");
 
+// Close the room when the game has ended
+exports.closeRoom = (roomId) => {
+    sockets.forEach(socket => socket.handshake.session.roomId === roomId && socket.leave(`room-${roomId}`));
```
jburton-scottlogic, line 101: lets say you had 100 rooms and the first one was the correct roomId, you would still be looping around the other 99 here. You could use 'find' to get the first instance in the array that matches.
```

const socket = sockets.find(socket => socket.handshake.session.roomId === roomId);
socket .leave(`room-${roomId}`));

```
or something similar

jburton-scottlogic: ah sorry i think i misunderstood, there will be multiple sockets that match the room id, ignore this

server/proxy/server.js
sibela46: I'm not really sure what you mean... this proxy server is ran only during testing mode though

server/services/socket.js
sibela46: Yeah, I think we're looping through the sockets here to make sure they leave their rooms when the game ends

server/proxy/server.js
jburton-scottlogic: ahhh! got you, nvm!

jburton-scottlogic (dismissed):

Ro4052 (approved):

tillywoodfield (approved):

PR #189, opened 2019-07-26
Displaying which players have finished their turn

sibela46:


server/services/GameLogic.js
```diff
@@ -239,6 +247,12 @@ module.exports = class GameLogic {
     /* Returns true if all players (apart from the current player) have voted this round */
     allPlayersVoted() { return this.players.every(player => player.username === this.currentPlayer.username || this.playerHasVoted(player.username)) };
 
+    /* Mark the player's turn as finished and notify the other players */
+    markTurnAsFinished() {
+        this.players.find(player => player.username === username).finishedTurn = true;
```
Ro4052, line 252: Where is `username` defined? Is it a global variable?

src/components/Players/Players.jsx
```diff
@@ -7,13 +7,16 @@ const cx = classNames.bind(styles);
 
 export class Players extends React.PureComponent {
     render() {
+        console.log(this.props.players);
```
Ro4052: Should probably remove this log.

server/services/GameLogic.js
sibela46: oops, that's when I moved it to a function, that's why the tests aren't passing, my bad

server/services/socket.js
```diff
@@ -9,6 +9,8 @@ exports.setupSocket = (server, session) => {
     io = socketio(server);
     io.use(sharedsession(session));
     io.on('connection', function (socket) {
+        console.log("hi");
```
Ro4052: Now remove this one?

sibela46: oops

Ro4052 (approved):

PR #212, opened 2019-07-29
Can logout or start a new game after finishing one

tillywoodfield:


server/controllers/index.js
```diff
@@ -41,11 +46,12 @@ router.post('/auth/logout', auth, (req, res) => {
...
         if (roomId !== null) {
             const gameState = getGameStateById(roomId);
-            // Why do we need both of these?
-            gameState.quitGame(user);
-            gameState.removePlayer(user);
-            if (gameState.getPlayers().length <= 0) games = games.filter(otherGame => otherGame !== gameState);
-            leaveRoom(user, roomId);
+            if (gameState) {
+                gameState.quitGame(user);
+                gameState.removePlayer(user);
```
sibela46: Do we need both of these?

tillywoodfield: I'm not sure, they were both there before so I left them in, will investigate whether one can be removed

sibela46 (approved):

PR #230, opened 2019-08-01
Dalmatian/timeout storyteller

tillywoodfield:


server/services/cardsManager.js
```diff
@@ -1,4 +1,4 @@
-module.exports.assign = (users, number_of_cards) => {
+exports.assign = (users, number_of_cards) => {
```
Ro4052: Not a fan of the underscore naming convention here, can it be changed to camel case?

src/components/Timer/TimerReducer.js
```diff
@@ -11,6 +12,8 @@ const timerReducer = (state = initialState, action) => {
             return {...state, playCardDuration: action.playCardDuration };
         case types.SET_VOTE_CARD_TIMER:
             return {...state, voteCardDuration: action.voteCardDuration };
+        case types.SET_STORYTELLER_TIMER:
+            return {...state, storytellerDuration: action.storytellerDuration };
```
Ro4052: Look at the inconsistent spacing! (and above) 😄

Ro4052 (commented):
Looks good, just a couple of comments.

src/components/Timer/TimerReducer.js
tillywoodfield: Inconsistent spacing?

Ro4052 (approved):

src/components/Dashboard/Dashboard.jsx
```diff
@@ -33,8 +33,9 @@ export class Dashboard extends React.Component {
                 </div>
                 <div className={styles.main}>
                     <div className={styles.side}>
-                        {(this.props.voteCard) && this.props.voteCardDuration && <Timer cy="vote-timer" reset={() => this.props.setVoteCardTimer(0)} duration={this.props.voteCardDuration} />}
-                        {(this.props.playCard) && this.props.playCardDuration && <Timer cy="card-timer" reset={() => this.props.setPlayCardTimer(0)} duration={this.props.playCardDuration} />}
+                        {(this.props.status === 'WAITING_FOR_CURRENT_PLAYER' && this.props.storytellerDuration > 0) && <Timer cy="storyteller-timer" reset={() => this.props.setStorytellerTimer(0)} duration={this.props.storytellerDuration} />}
```
jburton-scottlogic, line 36: Small comment - where you're using strings here 'WAITING_FOR_CURRENT_PLAYER' or 'STARTED' etc, these should probably be a static const that you import from somewhere.. e.g.
```
const DASH_STATUS = {
WAITING_FOR_CURRENT_PLAYER: 'WAITING_FOR_CURRENT_PLAYER',
STARTED: 'STARTED'
}
```
and then use it `this.props.status === DASH_STATUS.WAITING_FOR_CURRENT_PLAYER`

means you will consistently use the same string, no typo errors, if you want to add a new status everyone will know about it (by viewing the dash_status obj), changing the text of one of the statuses means you wont have to edit your whole code base etc etc.

jburton-scottlogic: Sorry just seen you already have it server/models/statusTypes.js but didn't use it here

tillywoodfield: Yeah we have it in there but React doesn't let you import things from outside of the `src` folder, would it be worth declaring the same object for the front end or is there a better way?

jburton-scottlogic: Sorry, been in meetings, hmm yeah fair point! There isn't really a clean way to do it, maintaining two copies (one for client, one for server) is probably better than using 'magic strings'. If the application back end was written in a different language you'd have to anyway

PR #245, opened 2019-08-02
Changed timeout duration and min players

tillywoodfield:


server/models/GameLogic.js
```diff
@@ -3,9 +3,10 @@ const socket = require('../services/socket');
 const { statusTypes } = require('./statusTypes');
 
 // Set durations
-const promptDuration = process.env.NODE_ENV === 'testing' ? 2000 : 20000;
+const promptDuration = process.env.NODE_ENV === 'testing' ? 2000 : 30000;
+const storytellerDuration = process.env.NODE_ENV === 'testing' ? 2000 : 60000;
 const nextRoundDuration = process.env.NODE_ENV === 'testing' ? 2000 : 5000;
-const minPlayers = process.env.NODE_ENV === 'testing' ? 2 : 2;
+const minPlayers = process.env.NODE_ENV === 'testing' ? 2 : 3;
```
Ro4052, line 9: In the future, if you want the min number of players to be 2 for development, I think you can do something like:
`const minPlayers = process.env.NODE_ENV === 'development' ? 2 : 3;`

Ro4052 (approved):

PR #249, opened 2019-08-02
Temporary fix for squished cards

tillywoodfield:


server/models/GameLogic.js
```diff
@@ -84,7 +84,7 @@ class GameLogic {
         } else if (this.players.some(player => player.username === username)) {
             throw Error("You have already joined this game");
         } else {
-            const cards = cardsManager.assign(this.players, 5);
+            const cards = cardsManager.assign(this.players, 4);
```
Ro4052, line 87: So are we just changing the number of cards a player has in their hand?

Ro4052: My review apparently didn't get submitted straight away.

tillywoodfield: We changed the css so that cards don't squish to long rectangles, but then also until we make a real fix for them not just going off the page, I thought reducing the number of cards in your hand from 5 to 4 would do. I guess this will be a job for the UX sprint next week

Ro4052: Was this the problem of it going of the bottom of the screen?

tillywoodfield: No the cards weren't card shaped anymore, they were like tall rectangles, so I set `flex-shrink: 0` and changed the height so they wouldn't do that, but then we had the problem of them being a constant size and if there were too many cards they just went off the page horizontally so we just got rid of one of the cards as a quick fix

Ro4052: As long as you think it won't have too much of an impact on game play. I'm a little hesitant but as long as it gets fixed in the next release...

tillywoodfield: In the last release we had 3 cards in a hand, so going up to 4 is still an improvement, and I'd rather have less cards than cards that don't look like cards

Ro4052 (approved):

PR #267, opened 2019-08-06
Eskipoo/speedup tests

tillywoodfield:


Ro4052 (approved):
Now you're the office Cypress/CircleCI expert! Although you might want to keep that hidden...

PR #301, opened 2019-08-09
Added a database to store scores

sibela46:


server/controllers/index.js
```diff
@@ -22,14 +29,28 @@ router.get('/auth', (req, res) => {
...
-        currentUsers.push({ username });
-        res.sendStatus(200);
-    }
+    db.get(username).then((user => {
+        if (!user) {
+            // TODO: Add this when we add authentication
+            // res.status(409).json({message: "Username already exists."});
+            try {
+                db.add(username).then((id) => {
+                    req.session.user = { username, id};
```
tillywoodfield: `req.session.user = { username, id };`

server/controllers/index.js
```diff
@@ -22,14 +29,28 @@ router.get('/auth', (req, res) => {
...
-        req.session.roomId = null;
-        currentUsers.push({ username });
-        res.sendStatus(200);
-    }
+    db.get(username).then((user => {
+        if (!user) {
+            // TODO: Add this when we add authentication
+            // res.status(409).json({message: "Username already exists."});
+            try {
+                db.add(username).then((id) => {
```
tillywoodfield: `db.add(username).then(id => {`

server/controllers/index.js
```diff
@@ -80,22 +102,26 @@ router.get('/api/start', auth, (req, res) => {
 router.get('/api/end', auth, (req, res) => {
     const { roomId } = req.session;
     try {
-        const gameState = Room.getById(req.session.roomId).gameState;
-        gameState.endGame();
-        Room.deleteById(roomId);
-        closeRoom(roomId);
-        res.sendStatus(200);
+        const gameState = Room.getById(roomId).gameState;
+        db.updateScores(gameState.getPlayers()).then((user) => {
```
tillywoodfield: `db.updateScores(gameState.getPlayers()).then(user => {`

tillywoodfield: Actually `db.updateScores(gameState.getPlayers()).then(() => {`?

server/controllers/index.js
```diff
@@ -80,22 +102,26 @@ router.get('/api/start', auth, (req, res) => {
...
         res.status(400).json({message: err.message});
     }
 });
 
 /* Current player plays a card and a word */
 router.post('/api/play-card-word', auth, (req, res) => {
-    if (Room.getById(req.session.roomId).gameState.isCurrentPlayer(req.session.user)) { /* Only current player is allowed to play both a word and a card */
+    const { user, roomId } = req.session;
+    const { word, cardId } = req.body;
+    if (Room.getById(roomId).gameState.isCurrentPlayer(user.username)) { /* Only current player is allowed to play both a word and a card */
```
tillywoodfield: We should probably move this validation into the GameLogic itself, seeing as we're doing try catch, same in the play-card request

sibela46: sure, should I do it in this PR?

server/controllers/index.js
sibela46: will be used eventually, but can remove it for now, yeah

server/services/socket.js
```diff
@@ -68,12 +68,18 @@ exports.closeRoom = roomId => {
...
 // Emit all the players
 exports.emitPlayers = (roomId, players) => sockets.forEach(socket => socket.handshake.session.roomId === roomId && socket.emit("players", { players }));
 
 // Let the players know about the next round
-exports.emitNewRound = (roomId, status, roundNum, currentPlayer) => sockets.forEach(socket => socket.handshake.session.roomId === roomId && socket.emit("new round", { status, roundNum, currentPlayer }));
+exports.emitNewRound = (roomId, status, roundNum, currentPlayer, timeoutDuration) => sockets.forEach(socket => {
+    if (socket.handshake.session.roomId === roomId) {
+        socket.emit("new round", { status, roundNum, currentPlayer });
+        if (socket.handshake.session.user.username === currentPlayer.username) {
+            socket.emit("play word and card", timeoutDuration/1000);
```
tillywoodfield, line 79: I guess we can remove promptCurrentPlayer if we're doing it here?

server/queries/queries.js
```diff
@@ -0,0 +1,104 @@
...
+    (err) => {
+    if (err) {
+        console.log(err);
+    } else {
+        console.log("Created table users");
+    }
+});
+
+// Add a user to the table
+module.exports.add = (username) => {
```
tillywoodfield: `module.exports.add = username => {` and in a bunch of other places in this file

server/queries/queries.js
```diff
@@ -0,0 +1,104 @@
...
+                })
+            } else {
+                resolve(result.rows[0]);
+            }
+        });
+    });
+}
+
+// Add a user to the table
+module.exports.updateScores = (players) => {
```
tillywoodfield: This function is getting a bit complicated, might look nicer if it was just `updateScore` and you did the looping outside?

server/queries/queries.js
```diff
@@ -0,0 +1,104 @@
...
+    return new Promise((resolve, reject) => {
+        let query = 'INSERT INTO users (name, score) VALUES ($1, $2) RETURNING id';
+        pool.query(query, [username, 0], (err, result) => {
+            if (err) {
+                reject({
+                    code: 500,
+                    msg: err
+                })
+            } else {
+                var id = result.rows[0].id;
```
tillywoodfield: const

server/queries/queries.js
```diff
@@ -0,0 +1,104 @@
...
+                resolve(id);
+            }
+        });
+    });
+}
+
+// Get user by username
+module.exports.get = (username) => {
+    return new Promise((resolve, reject) => {
+        let query = 'SELECT * FROM users WHERE name = $1';
```
tillywoodfield: const

server/queries/queries.js
sibela46: If I were to move it outside, it would be in controllers/index.js and I thought we should keep the logic outside this file

server/queries/queries.js
tillywoodfield: and in all the other queries

tillywoodfield: also some are called `query` and some are called `getUser` for example

server/queries/queries.js
```diff
@@ -0,0 +1,104 @@
...
+            }
+        });
+    });
+}
+
+// Add a user to the table
+module.exports.updateScores = (players) => {
+    return new Promise((resolve, reject) => {
+        players.forEach(player => {            
+            let getUser = 'SELECT * FROM users WHERE name = $1';
```
tillywoodfield: Can't this be done in a single query?

server/queries/queries.js
```diff
@@ -0,0 +1,104 @@
...
+    if (err) {
+        console.error('Connection error', err.stack);
+    } else {
+        console.log("Connected");
+    }
+});
+
+// Create the table if it doesn't already exist
+/* eslint max-len: ["error", { "code": 200 }] */
+const createTableQuery = 'CREATE TABLE IF NOT EXISTS users ( id SERIAL PRIMARY KEY, name VARCHAR(15) NOT NULL, score INT );'
```
scottlogic-alex: I think there's a more declarative way to do this.
I think Tilly declared all the domain entities and did `sequelize.sync()`. it was easier to guarantee idempotence and support incremental changes.

sibela46: Should I change the whole file to use Sequelize instead of Pool then?

server/queries/queries.js
sibela46: It can, I was just being silly

server/queries/index.js
```diff
@@ -0,0 +1,46 @@
...
+        logging:  false
+    })
+} else {
+    sequelize = new Sequelize(config.database, config.username, config.password, config);
+}
+
+fs
+  .readdirSync(__dirname)
+  .filter(file => {
+        return (file.indexOf('.') !== 0) && (file !== basename) && (file.slice(-3) === '.js');
```
Ro4052: Could have an implicit return here

server/controllers/index.js
```diff
@@ -72,72 +94,86 @@ router.get('/api/start', auth, (req, res) => {
...
     }
 });
 
 /* Player votes for a card */
-router.post('/api/vote-card', auth, (req, res) => { 
-    if (!Room.getById(req.session.roomId).gameState.isCurrentPlayer(req.session.user)) { /* Any user apart from the current player is allowed to vote for a card */
+router.post('/api/vote-card', auth, (req, res) => {
+    const { user, roomId } = req.session;
+    const { cardId } = req.body;
+    if (!Room.getById(roomId).gameState.isCurrentPlayer(user.username)) { /* Any user apart from the current player is allowed to vote for a card */
```
tillywoodfield: Can remove this one as well, and put the validation into game logic

server/queries/user.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+module.exports = function(sequelize, DataTypes) {
+
```
tillywoodfield: All the extra lines in this file look a bit weird

server/controllers/testqueries.js
```diff
@@ -0,0 +1,27 @@
...
+        nextId++;
+        users.push(user);
+        const data = {
+            dataValues: user
+        }
+        resolve([data]);
+    });
+}
+
+module.exports.update = () => {
```
tillywoodfield: What is this supposed to do?

sibela46: Update the score, but in the testqueries it does nothing for now as there aren't any tests for the leaderboard yet, I'll make them after I create the component

server/controllers/index.js
```diff
@@ -72,72 +94,86 @@ router.get('/api/start', auth, (req, res) => {
...
 router.get('/api/end', auth, (req, res) => {
     const { roomId } = req.session;
     try {
-        const gameState = Room.getById(req.session.roomId).gameState;
-        gameState.endGame();
-        Room.deleteById(roomId);
-        closeRoom(roomId);
-        res.sendStatus(200);
+        const gameState = Room.getById(roomId).gameState;        
+        gameState.getPlayers().forEach(player => {
```
tillywoodfield, line 94: Is this a sequelize thing? I'm a bit confused

sibela46: No that was there before, Room is a file you defined I think haha i'm getting a bit confused too

server/queries/index.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
```
tillywoodfield: This whole file seems really overly complicated? When we did it for the todo app I followed something only that is way simpler and more clear what's going on, if you want to see it

server/controllers/index.js
tillywoodfield: Oh I wasn't talking about Room I meant the stuff under this line sorry

sibela46: it was a sequelize thing yeah, but it's now different and hopefully clearer

server/controllers/index.js
```diff
@@ -22,14 +29,16 @@ router.get('/auth', (req, res) => {
 /* Log in the user */
 router.post('/auth/login', (req, res) => {
     const { username } = req.body;
-    if (currentUsers.find((user) => user.username === username)) {
-        res.status(409).json({message: "Username already exists."});
-    } else {
-        req.session.user = username;
+    db.createUser(username)
+    .then(users => {
+        const id = users[0].dataValues.id;
```
tillywoodfield, line 34: I think in general a create would return the new user or the id of the new user? Not `users`

server/controllers/index.js
```diff
@@ -72,72 +82,77 @@ router.get('/api/start', auth, (req, res) => {
...
     const { roomId } = req.session;
     try {
-        const gameState = Room.getById(req.session.roomId).gameState;
-        gameState.endGame();
-        Room.deleteById(roomId);
-        closeRoom(roomId);
-        res.sendStatus(200);
+        const gameState = Room.getById(roomId).gameState;        
+        gameState.getPlayers().forEach(player => {
+            db.updateScore(player.id)
```
tillywoodfield: How is it updating the score if not given the score?

server/controllers/index.js
sibela46: I think findOrCreate returns a list because it checks if a user with that username already exists and if it does, it returns that

server/controllers/index.js
sibela46: good point I'm updating it with the prev score, let me fix that

Ro4052 (approved):

PR #340, opened 2019-08-15
Flattie/next round button

sibela46:


src/components/PlayerInteractions/PlayerInteractions.jsx
```diff
@@ -4,28 +4,35 @@ import Prompt from '../shared/Prompt/Prompt';
...
 
 const cx = classNames.bind(styles);
 
 export class PlayerInteractions extends React.Component {
+    constructor(props) {
+        super(props);
+        this.requestNextRound = this.requestNextRound.bind(this);
+    }
+
+    requestNextRound() {
```
Ro4052: Not sure there is any need for this method, you could just call `this.props.requestNextRound` directly.

src/components/PlayerInteractions/PlayerInteractions.jsx
```diff
@@ -4,28 +4,35 @@ import Prompt from '../shared/Prompt/Prompt';
...
+        this.props.requestNextRound();
```
tillywoodfield: This doesn't need to be a whole function, and you can then remove the constructor right?

tillywoodfield: Oh I just realised Ronan already said this nevermind

tillywoodfield:
Maybe I've missed something but it looks like you might run into problems with the timeout on the server? If you press next round and then the timeout also goes next round?

sibela46:
> Maybe I've missed something but it looks like you might run into problems with the timeout on the server? If you press next round and then the timeout also goes next round?

The timeout is reset on every new round though, right?

src/components/Dashboard/Dashboard.jsx
```diff
@@ -31,13 +31,16 @@ export class Dashboard extends React.Component {
                         <Chat />
                     </div>
                     <div className={styles.middle}>
-                        <div className={styles.interactions}>
+                        {<div className={styles.interactions}>
```
tillywoodfield: What's this curly brace for?

sibela46: Nothing anymore 🙈

tillywoodfield (approved):

PR #342, opened 2019-08-15
Added hiding of chat

sibela46:


src/components/Chat/Chat.jsx
```diff
@@ -2,15 +2,36 @@ import React from 'react';
...
+        window.addEventListener('resize', this.updateWindowDimensions);
+    }
+      
+    componentWillUnmount() {
+        window.removeEventListener('resize', this.updateWindowDimensions);
+    }
+      
+    updateWindowDimensions() {
+        this.setState({
+            showChat: window.innerWidth > 950
```
tillywoodfield: I'm not that keen on this way of doing it, as it then hides the chat everytime you change the screen. Can't you use css media queries?

tillywoodfield (approved):

PR #347, opened 2019-08-16
Moved retrieve game state from endpoint to socket connection

tillywoodfield:


server/models/GameLogic.js
```diff
@@ -52,7 +54,7 @@ class GameLogic {
             currentPlayer: { username: this.currentPlayer.username }
         }
     }
-
+
```
sibela46: 👀

sibela46 (approved):

PR #371, opened 2019-08-20
Added basic github authentication

sibela46:


server/controllers/index.js
```diff
@@ -7,12 +7,45 @@ const { emitRooms, disconnectSocket, closeRoom, disconnectAllSockets } = require
 let currentUsers = [];
 let db;
 
+const oauthClientId = process.env.CLIENT_ID;
+const oauthSecret = process.env.CLIENT_SECRET;
+const redirectURL = process.env.NODE_ENV === 'development' ? "http://localhost:3000/lobby" : "/lobby";
```
Ro4052: I still think we shouldn't have to do this but I'm not sure what else to suggest. The only other thing I can think of is to remove the `proxy` from the `package.json` and bring back `setupProxy.js` with a new proxy for the oauth end point. I'm not convinced it's going to work but you can give it a go if you like.

src/components/Login/Login.jsx
```diff
@@ -21,6 +21,7 @@ export class Login extends React.Component {
 
     componentDidMount() {
         this.props.authenticateUser();
+        this.props.getURI();
```
Ro4052: Instead of calling this, can a failed `authenticateUser` call just return the URI?

server/controllers/index.js
sibela46: I'll give it a go

sibela46: It didn't work

server/controllers/oauth-github.js
```diff
@@ -0,0 +1,40 @@
...
+                });
+            }
+        });
+    }
+
+    return {
+      authorise: authorise,
+      oAuthUri: oAuthUri
+    };
+};
```
tillywoodfield: new line

server/controllers/oauth-github.js
```diff
@@ -0,0 +1,40 @@
...
+                }, (err, res, bd) => {
+                    const githubUser = JSON.parse(bd);
+                    callback(githubUser, token);
+                });
+            }
+        });
+    }
+
+    return {
+      authorise: authorise,
```
tillywoodfield, line 37: You can simplify this object and remove the colons

tillywoodfield (approved):

PR #373, opened 2019-08-20
Added our own authentication

sibela46:


server/controllers/index.js
```diff
@@ -58,25 +58,38 @@ router.get('/auth', (req, res) => {
...
+    db.createUser(username, password)
+    .then(user => {
+        const real = true;
+        const id = user.dataValues.id;
+        req.session.user = { username, id, real };
+        req.session.roomId = null;
+        currentUsers.push({ username });
+        res.sendStatus(200);
+    }).catch(err => {
+        console.log(err);
```
scottlogic-alex: `console.error()` is preferred, so you can filter by verbosity. you may also find that you enjoy better support in bash (i.e. errors sent to a separate file descriptor, stderr, which is not fed into pipes)

server/queries/db.js
```diff
@@ -17,23 +18,98 @@ sequelize.sync({ alter: true }).then(() => {
     console.log("Database connected.");
 });
 
-module.exports.createUser = username => {
+module.exports.validatePassword = (username, password) => {
+    return new Promise((resolve, reject) => {
+        User.findOne({
+            where: {
+                username: username
```
scottlogic-alex: `username: username` can be shortened to `username`

server/queries/db.js
```diff
@@ -17,23 +18,98 @@ sequelize.sync({ alter: true }).then(() => {
...
+            }
+        })
+        .then(user => {
+            if (user) {
+                bcrypt.compare(password, user.dataValues.password, (err, res) => {                    
+                    if (res) {
+                        resolve(user);
+                    } else {
+                        reject({
+                            code: 404,
```
scottlogic-alex: `/auth/login` doesn't appear to use these codes

scottlogic-alex: are you intending to describe an HTTP status code?

404 would be inappropriate. the resource `/auth/login` was found.

401 Unauthorized would be appropriate for a failed authentication attempt.

scottlogic-alex: worth considering 200 OK as in "there were no errors using the API; you used the endpoint correctly." but the body could say "`{success: false}`"

sibela46: why would the success be false?

server/queries/db.js
```diff
@@ -17,23 +18,98 @@ sequelize.sync({ alter: true }).then(() => {
...
+                            message: "Password is incorrect"
+                        });
+                    }
+                });
+            } else {
+                reject({
+                    code: 404,
+                    message: "User with this username does not exist"
```
scottlogic-alex, line 41: sometimes it's considered bad to reveal whether a user exists. you have to decide whether you would prefer usability, or privacy+security.

server/queries/db.js
```diff
@@ -17,23 +18,98 @@ sequelize.sync({ alter: true }).then(() => {
...
+                    code: 404,
+                    message: err.message
+                });
+            }
+            User.findOrCreate({
+                where: {
+                    username: username
+                },
+                defaults: { // set the default properties if it doesn't exist
+                    username: username.trim(),
```
scottlogic-alex: what's with the `trim()`? is this defensive programming? is there a specific problem you are trying to guard against?

scottlogic-alex: it's better to tell the user that their input is invalid, rather than "fix" it without communicating to them.

sibela46: No, they were just using it at one tutorial, so I followed that

server/queries/db.js
```diff
@@ -17,23 +18,98 @@ sequelize.sync({ alter: true }).then(() => {
...
+        });
+    });
+}
+
+module.exports.createUser = (username, password) => {
+    return new Promise((resolve, reject) => {
+        bcrypt.hash(password, 10, (err, hash) => {
+            if (err) {
+                reject({
+                    code: 404,
```
scottlogic-alex: my same feedback concerning `code` applies to this function, too

server/queries/db.js
```diff
@@ -17,23 +18,98 @@ sequelize.sync({ alter: true }).then(() => {
...
+                    username: username
+                },
+                defaults: { // set the default properties if it doesn't exist
+                    username: username.trim(),
+                    password: hash,
+                    score: 0
+                }
+            })
+            .then(([user, created]) => {                
+                if (!created) {
```
scottlogic-alex: this code would read better if you inverted the condition -- otherwise we get a double-negative when we do the else

server/queries/db.js
sibela46: usability

server/queries/db.js
```diff
@@ -17,23 +18,94 @@ sequelize.sync({ alter: true }).then(() => {
...
+                where: { username },
+                defaults: { // set the default properties if it doesn't exist
+                    username,
+                    password: hash,
+                    score: 0
+                }
+            })
+            .then(([user, created]) => {                
+                if (!created) {
+                    reject({
```
scottlogic-alex: if you `reject(); return` then you don't need the else, because flow ends inside the conditional block.

src/components/Login/Login.jsx
```diff
@@ -4,46 +4,63 @@ import { Redirect } from 'react-router-dom';
...
-                        <h2 className={styles.formHeader}>Type a username to enter the game:</h2>
-                        <input data-cy='username' className={cx(styles.loginInput, { inputError: this.props.error })} value={this.state.value} placeholder="Enter username" onChange={this.handleChange} autoFocus />
+                    <form className={styles.loginForm} onSubmit={this.sendLoginOrSignUp} data-cy='login-form' >
+                        <h2 className={styles.formHeader}>{this.state.signUp ? "Enter a username and password to sign up for the game" : "Log in with your username and password to enter the game"}</h2>
+                        <input type='text' data-cy='username' className={cx(styles.loginInput, { inputError: this.props.error })} username={this.state.username} placeholder="Enter username" onChange={this.handleChangeUsername} autoFocus />
+                        <input type='password' data-cy='password' className={cx(styles.loginInput, { inputError: this.props.error })} password={this.state.password} placeholder="Enter password" onChange={this.handleChangePassword} autoFocus />
```
scottlogic-alex, line 56: there was a famous password leak that occurred recently, because React was managing the data inside a `<input type="password"`. I've been unable to find the article, but I wonder whether it's safe to bind `password` to `this.state`. I think to secure it, you need to ensure that javascript doesn't touch sensitive data.

scottlogic-alex: ordinarily, I think javascript cannot see the value inside a `<input type="password">`, but this model-view binding circumvents the browser's security.

src/components/Login/LoginActions.js
```diff
@@ -71,14 +71,42 @@ export const getURI = () => dispatch => {
...
         if (!allowed.test(username)) throw Error("Username can be comprised of numbers and latin letters only");
-        axiosInstance.post('/auth/login', { username })
+        if (!allowed.test(password)) throw Error("Password can be comprised of numbers and latin letters only");
+        axiosInstance.post('/auth/login', { username, password })
+        .then(res => {
+            if (res.status === 200) {
+                connectSocket()
+                .then(() => dispatch(authSuccess(username)));
+            } else {
+                throw Error(res.data.message);
```
scottlogic-alex: presumably you can dispatch authFailure() here, instead of catching it and doing it in the promise chain?

src/components/Login/LoginActions.js
```diff
@@ -71,14 +71,42 @@ export const getURI = () => dispatch => {
...
+export const signUp = (username, password) => dispatch => {
+    try {
+        if (username.length < 3) throw Error("Username must be at least 3 characters");
+        if (username.length > 15) throw Error("Username must be no longer than 15 characters");
+        if (password.length < 3) throw Error("Password must be at least 3 characters");
+        if (password.length > 15) throw Error("Password must be no longer than 15 characters");
+        // eslint-disable-next-line
+        const allowed = /^[A-Za-z0-9]*$/;
+        if (!allowed.test(username)) throw Error("Username can be comprised of numbers and latin letters only");
+        if (!allowed.test(password)) throw Error("Password can be comprised of numbers and latin letters only");
```
scottlogic-alex: why are we trying to make passwords less secure?

src/components/Login/LoginActions.js
```diff
@@ -71,14 +71,42 @@ export const getURI = () => dispatch => {
...
+};
+
+export const signUp = (username, password) => dispatch => {
+    try {
+        if (username.length < 3) throw Error("Username must be at least 3 characters");
+        if (username.length > 15) throw Error("Username must be no longer than 15 characters");
+        if (password.length < 3) throw Error("Password must be at least 3 characters");
+        if (password.length > 15) throw Error("Password must be no longer than 15 characters");
+        // eslint-disable-next-line
+        const allowed = /^[A-Za-z0-9]*$/;
```
scottlogic-alex: I think `\w` (word) is a group that matches alphanumerics

src/components/Login/LoginActions.js
sibela46: that's a good point

cypress/support/commands.js
```diff
@@ -24,12 +24,13 @@
 // -- This is will overwrite an existing command --
 // Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
 
-Cypress.Commands.add('login', username => {
+Cypress.Commands.add('login', (username, password) => {
```
tillywoodfield: Can you change this command to `signUp`?

server/queries/testdb.js
```diff
@@ -10,18 +10,24 @@ module.exports.reset = () => {
...
             user = { username, id: nextId, score: 0 };
             nextId++;
             users.push(user);
+            const data = {
+                dataValues: user
+            }
+            resolve(data);
+        } else {
+            reject({
+                code: 404,
```
tillywoodfield: 404 isn't the correct code for a username already existing

src/components/Login/Login.jsx
```diff
@@ -4,46 +4,63 @@ import { Redirect } from 'react-router-dom';
...
-    sendLogin(event) {
+    handleChangePassword(event) {
+        this.setState({ password: event.target.value });
+    }
+
+    sendLoginOrSignUp(event) {
         event.preventDefault();
-        this.props.logIn(this.state.value);
+        console.log(this.state.username, this.state.password);
+
```
tillywoodfield: Console log and whitespace

src/components/Login/Login.jsx
tillywoodfield: How would you do it instead?

tillywoodfield (approved):

PR #387, opened 2019-08-23
Goldie/intelligence

iatkins:


server/models/GameLogic.js
```diff
@@ -361,9 +361,9 @@ class GameLogic {
 
     updateLabels(word) {
         this.playedCards.forEach(card => {
-            const cardScore = this.votes.filter(vote => card.cardId === vote.cardId).length
+            const cardScore = this.votes.filter(vote => card.cardId === vote.cardId && !vote.username.includes("CPU")).length;
```
tillywoodfield: Someone may have CPU in their username, is it better to find the user and use the `user.real` flag

server/queries/db.js
```diff
@@ -53,6 +53,29 @@ module.exports.validatePassword = (username, password) => {
...
+module.exports.addLabel = (cardId, word) => {
+    return new Promise((resolve, reject) => {
+        CardLabels.findOne({
+            where: { cardId }
+        }).then(card => {
+            const labels = card.dataValues.labels;
+            if (!labels.includes(word)) {
+                labels.push(word);
+                card.update({
+                labels: labels
```
tillywoodfield: indent, and `labels: labels` can be simplified to `labels`

server/services/AI/AI.js
```diff
@@ -53,3 +45,11 @@ exports.autoWord = (cardId) => {
         });
     });
 }
+
+exports.addLabel = (cardId, word) => {
+    return new Promise((resolve, reject) => {
+        db.addLabel(cardId, word);
+        resolve();
+
```
tillywoodfield: extra line

server/services/AI/AI.js
```diff
@@ -53,3 +45,11 @@ exports.autoWord = (cardId) => {
         });
     });
 }
+
+exports.addLabel = (cardId, word) => {
+    return new Promise((resolve, reject) => {
+        db.addLabel(cardId, word);
+        resolve();
```
tillywoodfield: This should be a `.then`

server/services/AI/AI.js
```diff
@@ -53,3 +45,11 @@ exports.autoWord = (cardId) => {
...
+
+    });
+}
```
tillywoodfield: new line

sibela46:
Do you wanna make it so that each label is stored together with its score (which is increased every time someone plays that word), so that when the AI picks a label it's not at random but based on a priority score

tillywoodfield (approved):

PR #396, opened 2019-08-27
made custom mode intelligent

iatkins:


server/services/rekognition/rekognition.js
```diff
@@ -0,0 +1,26 @@
+const AWS = require('aws-sdk');
+var creds = new AWS.Credentials(process.env.AWS_ACCESS_KEY, process.env.AWS_SECRET_KEY);
```
Ro4052: No `var` please 😄 As far as I can see, they could all be `const`.

server/services/rekognition/rekognition.js
```diff
@@ -0,0 +1,26 @@
...
+                Name: "0aa220f92872282ef7cc2b1800527e99.jpg"
+            }
+        }
+    };
+    rekognition.detectLabels(params, function (err, data) {
+        if (err) {
+            console.log(err, err.stack);
+        } else {
+            const labels = data.Labels.map(label => label.Name)
+            console.log(labels);
```
Ro4052: I'm not sure exactly what you're trying to log but you may find this dilutes the server logs a bit.

server/services/rekognition/rekognition.js
```diff
@@ -15,12 +15,16 @@ function assignLabels(imageData) {
...
-            console.log(labels);
-        }    
+    return new Promise((resolve, reject) => {
+        rekognition.detectLabels(params, function (err, data) {
+            if (err) {
+                console.log(err, err.stack);
+                resolve(err);
+            } else {
+                const labels = data.Labels.map(label => label.Name)
+                console.log(labels);
```
Ro4052: Might want to get rid of this log, up to you.

server/services/AI/AI.js
```diff
@@ -35,24 +35,25 @@ exports.autoWordGenerator = autoWordGenerator
...
-        if (mode === 'original') {
-            db.getLabels(cardId).then(card => {
-                const labels = card.dataValues.labels;
-                const word = labels[Math.floor(Math.random()*labels.length)];
-                resolve(word);
-            });
-        } else {
-            resolve("Word");
+exports.autoWord = (cardId, gameMode) => {
+    let promise;
```
Ro4052: Could turn this into a ternary:
```js
const promise = gameMode === 'original' ? db.getLabelsTellTales(cardId) : db.getLabelsCustomMode(cardId)
```

server/models/GameLogic.js
```diff
@@ -350,9 +350,7 @@ class GameLogic {
         this.newCard();
         this.update({ status: statusTypes.DISPLAY_ALL_VOTES });
         this.calcScores();        
-        if (this.state.mode === 'original') {
```
Ro4052, line 353: Did you really want to remove this if statement?

iatkins: so this if statement was because sisi was trying to avoid calling the update labels in the custom mode, because that wasn't set up yet, so she just wanted to skip the update if that makes sense. but now that i've made the update labels work for custom mode we don't need the if statement

Ro4052: Ah, cool 👍

Ro4052 (approved):
Guess this is the last approval from me 🕯

PR #397, opened 2019-08-28
Goldie/image upload

sibela46:


server/controllers/images.js
```diff
@@ -0,0 +1,56 @@
+const express = require('express');
+const aws = require('aws-sdk');
+const multer = require('multer');
+const multerS3 = require('multer-s3');
+const router = express.Router();
+const auth = require('../middlewares/auth');
+
+let db;
```
Ro4052: Just a thought, if you used a ternary here, you could make `db` a `const`.

sibela46: that's true... also just realised I don't need the other db here, because I can't simulate an image upload from cypress, so testing for that functionality will only happen by client tests.

server/models/room.js
```diff
@@ -38,12 +32,36 @@ const update = roomId => state => {
...
+                }
+            })
+            .catch(err => {
+                reject({
+                    code: err.code,
+                    message: err.message
+                });
+            });
+        } else {
+            const gameState = new GameLogic(update(roomId), roomId, numRounds, gameMode, 247 /*cards length*/);
```
EllieHield: comment left in the args here

sibela46: I put it on purpose, thought it would be more clear for someone reading the code, otherwise it's just a number 😄

sibela46: and if the number of original cards changes, it'd be easier to find where to change the size of the deck

server/controllers/images.js
```diff
@@ -0,0 +1,50 @@
...
+const singleUpload = upload.single('image');
+
+router.post("/upload", auth, (req, res) => {    
+    singleUpload(req, res, err => {
+        if(err) {
+            console.log(err);
+            return res.status(401).json({err});
+        }
+        db.addCard(req.file.etag, req.file.location)
+        .then((card) => {
```
Ro4052: Could have this if you like:
```js
.then(card => res.status(200).json({ 'imageUrl': card.url }))
```

server/models/room.js
EllieHield: oh sorry I thought it was a separate argument! Is it worth defining that as a const so the variable name speaks for itself?

sibela46: that's a good point, idk why I didn't think of that haha, thanks!

Ro4052: If you want clarity, maybe you should make it a variable (constant).

_Edit: Doing my review before I saw the other comments_

src/components/Lobby/CreateRoom/CreateRoom.jsx
```diff
@@ -11,18 +11,23 @@ export class CreateRoom extends React.Component {
...
         return (
             <div className={styles.createRoom}>  
                 <h1 className={styles.createRoomHeader}>Create room</h1>
+                <span>{this.props.error}</span>
                 <form className={styles.createRoomForm} data-cy="create-room-form" onSubmit={this.createRoom}>
                     <select className={styles.selectRounds} data-cy="num-rounds-options" onChange={event => this.props.setRoundCount(event.target.value)}>
                         {options.map((option, key) => <option key={key} value={option}>{option} rounds</option>)}
                     </select>
+                    <select className={styles.selectRounds} data-cy="mode-options" onChange={event => this.props.setGameMode(event.target.value)}>
+                        <option value={"telltales"}>TellTales Images</option>
```
Ro4052:
```suggestion
                        <option value="telltales">TellTales Images</option>
```

src/components/Lobby/CreateRoom/CreateRoom.jsx
```diff
@@ -11,18 +11,23 @@ export class CreateRoom extends React.Component {
...
+                        <option value={"custom"}>Custom Images</option>
```
Ro4052:
```suggestion
                        <option value="custom">Custom Images</option>
```

src/components/shared/CardList/Card/Card.jsx
```diff
@@ -26,7 +26,9 @@ export class Card extends React.Component {
     }
 
     render() {
-        const src = this.props.card.cardId ? require(`../../../../images/cards/card (${this.props.card.cardId}).jpg`) : require('../../../../images/cardBack.jpg');
+        let src;        
+        if (this.props.gameMode === 'telltales') src = this.props.card.cardId ? require(`../../../../images/cards/card (${this.props.card.cardId}).jpg`) : require('../../../../images/cardBack.jpg')
```
Ro4052: I think having a game mode called "telltales" is a bit confusing. Maybe something like "original"?

src/components/shared/CardList/Card/Card.jsx
```diff
@@ -26,7 +26,9 @@ export class Card extends React.Component {
...
+        else if (this.props.gameMode === 'custom') src=this.props.card.url;
```
EllieHield: might be nicer as a switch statement given there's defined options but that's down to your opinion?

EllieHield: I would suggest multi lining the ternaries so it's a bit easier to follow like:
statement
   ? option 1
   : option 2

sibela46: what did you have in mind?

sibela46: a switch on the game mode or a switch on the cardId

Ro4052: Or:
```js
const src = {
  telltales: this.props.card.cardId ? require(`../../../../images/cards/card (${this.props.card.cardId}).jpg`) : require('../../../../images/cardBack.jpg'),
  custom: src=this.props.card.url
}[this.props.gameMode]
```

EllieHield: something like 
```
switch(this.props.gameMode) {
    case "telltales" 
         ternary...
         break
    case "custom"
         other thing...
         break
}
```
But Ronan's thing is also cool, it was just a clarity thing since you had nested ifs and ternaries

Ro4052: I only suggested that other option because it's something we do quite a lot on my project.

server/controllers/images.js
scottlogic-alex: quotes around `'imageUrl'` are unnecessary

server/controllers/images.js
```diff
@@ -0,0 +1,50 @@
...
+    secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY,
+    accessKeyId: process.env.AWS_ACCESS_KEY_ID,
+    region: 'eu-west-2'
+});
+
+const s3 = new aws.S3();
+
+const upload = multer({
+    storage: multerS3({
+        s3: s3,
```
scottlogic-alex: `s3: s3,` can be shortened to `s3,`

server/controllers/index.js
```diff
@@ -120,6 +115,15 @@ router.get('/api/cards', auth, (req, res) => {
     const gameState = Room.getById(roomId).gameState;
     const cards = gameState.getUnplayedCardsByUsername(user.username);
     if (cards) {
+        if (gameState.mode === 'custom') {
+            cards.forEach(card => {
```
scottlogic-alex: this is capable of invoking `res.status(err.code)` n times. it's not possible to respond multiple times to the same HTTP request.

moreover, there's a race condition. when game mode is custom, it will fire off all these async functions, but it will also continue to execute `res.status(200)`.

scottlogic-alex: you should do something like:

```js
(gameState.mode === 'custom'
? Promise.all(
  cards.map(card => db.getCard(card.cardId)
    .then(other => ({
      ...card,
      url: other.dataValues.url
    }))))
: Promise.resolve(cards))
.then(
  cards => res.status(200).json(cards),
  err => res.status(500).json({ message: err.message ));
```

scottlogic-alex: also we shouldn't use err.code as our HTTP status code.  
I am assuming `db.getCard()` would return _database_ error codes, not HTTP status codes. and even if it happened to be HTTP: the client talking to our backend wants to know the status of our _backend_, not of other services it depends on.

server/controllers/rooms.js
```diff
@@ -19,13 +19,19 @@ router.post('/create', auth, (req, res) => {
...
+        .then(newRoomId => {
+            const newRoom = Room.getById(newRoomId);
+            Room.addPlayer(newRoom, user);
+            socket.joinRoom(newRoomId, user.username);
+            socket.emitRooms();
+            req.session.roomId = newRoomId;                    
+            res.sendStatus(200);
+        })
+        .catch(err => {
+            console.log(err)
```
scottlogic-alex:
```js
console.error(err)
```

server/models/GameLogic.js
```diff
@@ -14,11 +14,13 @@ exports.minPlayers = minPlayers;
...
 
 class GameLogic {
-    constructor(notifyObserver, roomId, numRounds) {
+    constructor(notifyObserver, roomId, numRounds, gameMode, allCardsNumber) {
         this.notifyObserver = notifyObserver;
         this.state = {
             status: statusTypes.NOT_STARTED,
             roundNum: 0,
+            mode: gameMode,
+            allCardsNumber: allCardsNumber,
```
scottlogic-alex: shorthand:  
`allCardsNumber,`

server/models/GameLogic.js
```diff
@@ -85,7 +88,7 @@ class GameLogic {
     /* Return the list of cards played this round, hiding who played them */
     getPlayedCards() {
         if (this.state.status === statusTypes.WAITING_FOR_VOTES) {
-            return this.state.playedCards.map(card => ({ cardId: card.cardId }));
+            return this.state.playedCards.map(card => ({ cardId: card.cardId, url: card.url }));
```
scottlogic-alex: shorthand using destructuring:
```js
.map(({ cardId, url }) => ({ cardId, url }))
```
or just don't map it at all :P

iatkins (dismissed):

server/controllers/images.js
```diff
@@ -0,0 +1,48 @@
...
+router.post("/upload", auth, (req, res) => {    
+    singleUpload(req, res, err => {
+        if(err) {
+            console.log(err);
+            return res.status(401).json({err});
+        }
+        db.addCard(req.file.etag, req.file.location)
+        .then(card => res.status(200).json({ imageUrl: card.url }))
+        .catch(err => {
+            res.status(err.code).json({ message: err.message });
```
scottlogic-alex: why repeat the database error code to the client? is it even an HTTP status code?

I think it's more appropriate to return HTTP 500 (our backend encountered an internal error)

server/controllers/index.js
```diff
@@ -4,19 +4,13 @@ const auth = require('../middlewares/auth');
...
 const redirectURL = process.env.NODE_ENV === 'production' ? "/lobby" : "http://localhost:3000/lobby";
 const oAuthGithub = require('./oauth-github');
 const githubAuthoriser = oAuthGithub(oauthClientId, oauthSecret);
 
-if (process.env.NODE_ENV === 'testing') {
-    db = require('../queries/testdb');
-} else {
-    db = require('../queries/db');
-}
+const db = (process.env.NODE_ENV === 'testing') ? require('../queries/testdb') : require('../queries/db');
```
scottlogic-alex: this ternary can be inside the `require()`

```js
const db = require(`../queries/${
  process.env.NODE_ENV === 'testing'
  ? 'testdb'
  : 'db'
}`);
```

server/controllers/rooms.js
```diff
@@ -19,15 +19,21 @@ router.post('/create', auth, (req, res) => {
...
+            const newRoom = Room.getById(newRoomId);
+            Room.addPlayer(newRoom, user);
+            socket.joinRoom(newRoomId, user.username);
+            socket.emitRooms();
+            req.session.roomId = newRoomId;                    
+            res.sendStatus(200);
+        })
+        .catch(err => {
+            console.error(err)
+            res.status(400).json({ message: err.message });
```
scottlogic-alex, line 33: why HTTP 400? did the client send you an invalid request?

server/models/room.js
```diff
@@ -38,12 +32,37 @@ const update = roomId => state => {
...
+exports.create = (numRounds, gameMode) => {
+    return new Promise((resolve, reject) => {
+        const roomId = latestRoomId;
+        latestRoomId++;
+        if (gameMode === 'custom') {
+            db.getAllCards()
+            .then(cards => {                         
+                if (cards.length >= 50) {
+                    const gameState = new GameLogic(update(roomId), roomId, numRounds, gameMode, cards.length);
+                    rooms.push({ roomId, title: roomNames[Math.floor(Math.random() * 78)], gameState });
```
scottlogic-alex: why 78? do you mean `roomNames.length`?

server/queries/db.js
```diff
@@ -76,6 +78,57 @@ module.exports.addLabel = (cardId, word) => {
     });
 }
 
+module.exports.addCard = (etag, url) => {
+    return new Promise((resolve, reject) => {
```
scottlogic-alex: this is an unnecessary promise.
`CardImages.findOrCreate()` already returns a thenable, so you should chain from that

```js
(etag, url) =>
CardImages.findOrCreate(/* whatever */)
.then(([card, created]) => {
  if (created) return card;
  throw {
    code: 409,
    message: "Card already exists in the database"
  };
})
```

src/components/Upload/Upload.jsx
```diff
@@ -0,0 +1,57 @@
...
+        this.sendImageUpload = this.sendImageUpload.bind(this);
+        this.onChange = this.onChange.bind(this);
+    }
+
+    sendImageUpload(e){
+        e.preventDefault();
+        if (!this.state.file) {
+            this.props.uploadImageFailure("You need to select an image first.");
+            this.setState({ file: null })
+        } else if (this.state.file && (this.state.file.type === 'image/jpeg' || this.state.file.type === 'image/png')) {
```
scottlogic-alex:
```js
['image/jpeg', 'image/png'].includes(this.state.file.type)
```

src/components/Upload/UploadActions.js
```diff
@@ -0,0 +1,32 @@
...
+        headers: {
+            'content-type': 'multipart/form-data'
+        }
+    };
+    axiosInstance.post("/api/images/upload", formData, config)
+    .then(res => {
+        if (res.status === 200) {
+            dispatch(uploadImageSuccess("File was uploaded successfully!"));
+        } else {
+            dispatch(uploadImageFailure(res.data.message));
```
scottlogic-alex: you could re-use your existing `.catch()` by throwing here:

```js
throw new Error(res.data.message);
```

server/queries/db.js
sibela46: but then I want addCard to return a thenable too

iatkins (approved):

PR #399, opened 2019-08-29
Hopefully fixed bug

sibela46:


server/models/GameLogic.js
```diff
@@ -117,8 +119,19 @@ class GameLogic {
         } else if (this.state.players.length === maxPlayers) {
             throw Error("The room has reached its capacity");
         } else {
-            const cards = cardsManager.assign(this.state.players, 6, this.state.allCardsNumber)
-            const player = { username: user.username, id: user.id, cards, score: 0, real: user.real, finishedTurn: false };
+            const cards = cardsManager.assign(this.state.players, 6, this.state.allCardsNumber);
+            let cardsWithUrls = [];
```
Ro4052: Pretty sure it can be a `const`
```suggestion
            const cardsWithUrls = [];
```

server/models/GameLogic.js
```diff
@@ -354,7 +377,7 @@ class GameLogic {
         this.update();
     }
 
-    updateLabels(word) {
+    updateLabels(word) {
```
Ro4052: Some extra white space.

server/models/GameLogic.js
sibela46: but then you can't alter it?

Ro4052: But you're only pushing onto it right? Making it a `const` just means you can't re-assign (`cardsWithUrls = someOtherArray`), i.e. you can't change the reference.

Ro4052 (approved):

PR #401, opened 2019-08-30
fix story teller timeout value

Ro4052:


sibela46 (approved):
This is an absolute piece of art.

PR #411, opened 2019-09-02
Hide remove AI button when not in room or game has started

tillywoodfield:


Ro4052 (approved):

src/components/Lobby/Room/Room.jsx
```diff
@@ -6,26 +6,31 @@ import { startGame, joinRoom, leaveRoom, addAIPlayer, removeAIPlayer } from '../
...
-                    {!this.props.room.started && this.props.room.players.length !== this.props.room.maxPlayers && !inRoom && <button className={styles.roomButton} onClick={() => this.props.joinRoom(this.props.room.roomId)} data-cy="join-room" type='button'>Join room</button>}
-                    {!this.props.room.started && this.props.room.players.length !== this.props.room.maxPlayers && inRoom && <button className={styles.roomButton} data-cy="automated-player" onClick={() => this.props.addAIPlayer(this.props.room.roomId)} >Add AI player</button>}
-                </div>
+            <div className={styles.room} data-cy="room">
+                <h2 className={styles.title} data-cy="room-title">{`Room: ${this.props.room.title}`}</h2>
+                <table className={styles.players}>
+                    <tbody data-cy='room-players'>
+                        {this.props.room.players.map((player, key) => (
+                            <tr className={styles.player} key={key}>
```
Ro4052, line 18: My only point is that I would avoid using an index as a key but I don't think it's a huge deal.

tillywoodfield: I think we might've had it as player id before but the AI players all have an id of 0 so it wouldn't work

PR #415, opened 2019-09-02
Made only arrow area clickable

tillywoodfield:


src/components/Chat/Chat.jsx
```diff
@@ -44,9 +44,11 @@ export class Chat extends React.Component {
     render() {
         return (
             <>
-                <div className={styles.chatArrowArea} onClick={this.toggleChat}>
-                    <div className={cx({ showChatArrow: !this.state.showChat, hideChatArrow: this.state.showChat })} data-cy='chat-arrow'/>
-                    <span onClick={this.showChat} data-cy='show-chat'>{this.state.showChat ? "" : "Chat"}</span>
+                <div className={styles.chatArrowArea}>
+                    <div className={styles.clickable} onClick={this.toggleChat}>
+                        <div className={cx({ showChatArrow: !this.state.showChat, hideChatArrow: this.state.showChat })} data-cy='chat-arrow'/>
+                        <span onClick={this.showChat} data-cy='show-chat'>{this.state.showChat ? "" : "Chat"}</span>
```
Ro4052: Why do we have a span with no text rather than just removing it?

tillywoodfield: I don't know, will change now

Ro4052 (approved):

PR #416, opened 2019-09-02
Refactored lobby page to stop weird scrolling and make the layout more responsive

tillywoodfield:


src/components/Lobby/Header/Header.jsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import styles from './Header.module.css';
+import Logo from '../../Logo/Logo';
+import Logout from '../../Logout/Logout';
+import LeaderboardButton from '../../Leaderboard/LeaderboardButton/LeaderboardButton';
+import UploadButton from '../../Upload/UploadButton/UploadButton';
+
+export class Header extends React.Component {
```
Ro4052: Might as well make this a functional component.

src/components/Upload/Upload.jsx
```diff
@@ -16,6 +17,10 @@ export class Upload extends React.Component {
         this.onChange = this.onChange.bind(this);
     }
 
+    goToLobby() {
+        history.push('/lobby');
```
Ro4052: I thought we'd gone with the approach of using the `history` object passed down through the tree?

Ro4052 (dismissed):

src/components/Upload/Upload.jsx
tillywoodfield: I just copied this function from the leaderboard page, but happy to change it?

Ro4052: That's up to you. Another point in a similar area though, is the `UploadButton` class ever being used?

src/components/Upload/UploadButton/UploadButton.jsx
```diff
@@ -1,18 +1,18 @@
 import React from 'react';
-import history from '../../../services/history';
+import { withRouter } from "react-router";
 
 export class UploadButton extends React.Component {
     goToUpload() {
-        history.push('/upload');
+
```
Ro4052: Added an extra line, also, is this file actually used?

Ro4052 (dismissed):

Ro4052:
I keep accidently clicking approve instead of comment. One of the tests is failing anyway...

src/components/Upload/UploadButton/UploadButton.jsx
tillywoodfield: The history file is still used in the redux actions

Ro4052 (approved):
This time I _meant_ to approve.

